PR #47, opened 2021-09-15
Add support for external processes

Jaswar:


src/main/java/nl/tudelft/cse1110/andy/execution/process/ExternalProcess.java
```diff
@@ -0,0 +1,79 @@
+package nl.tudelft.cse1110.andy.execution.process;
+
+import java.io.IOException;
+import java.util.Scanner;
+
+public class ExternalProcess extends Thread {
```
mauricioaniche, line 6: It's usually more common to implement `Runnable` rather than extending `Thread`, but I guess the final result is the same.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunExternalProcessStep.java
```diff
@@ -0,0 +1,39 @@
...
+
+            if (externalProcess == null) {
+                return;
+            }
+
+            externalProcess.launch();
+
+            ctx.setExternalProcess(externalProcess);
+
+            externalProcess.await();
```
mauricioaniche, line 28: Is this call blocking? I can't guess it by the implementation of `await`. If so, this is a problem. We want the rest of the pipeline to work while the other process is executing.

src/main/java/nl/tudelft/cse1110/andy/result/Result.java
```diff
@@ -14,8 +14,9 @@
     private final String genericFailure;
     private final double timeInSeconds;
     private final GradeWeight weights;
+    private final String externalProcessOutput;
```
mauricioaniche, line 17: We do not care about the external process output, unless it's an error, in which case we just set is as "genericFailure"

src/test/java/integration/ExternalProcessTest.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Test
+    void externalProcessGracefulExit() {
+        Result result = run("EmptyLibrary", "EmptySolution",
+                "ExternalProcessGracefulExitConfiguration");
+
+        assertThat(result.getGenericFailure()).isNull();
+
+        assertThat(result.getExternalProcessOutput()).isEqualTo("hello\n");
```
mauricioaniche, line 58: Indeed, no need to assert the result of the external process. What we can do here is to assert that the rest of the pipeline worked as expected.

src/test/resources/grader/fixtures/Config/ExternalProcessEndSignalConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+            put("meta", 1.0f);
+            put("codechecks", 0.0f);
+        }};
+    }
+
+    @Override
+    public ExternalProcess externalProcess() {
+        String tmp = System.getProperty("java.io.tmpdir");
+        return new ExternalProcess(
+                "sh " + tmp + "/andy_test_external_process_end_signal.sh",
```
mauricioaniche, line 32: Oh that's a good thing to consider. The script will also pass the directory in which the sh is, ok, that should work!

src/test/java/integration/ExternalProcessTest.java
mauricioaniche: One challenge would be to indeed ensure that the script was executed. Maybe the sh file can simply create a file in a temp directory and our test checks the existence of this file?

mauricioaniche (commented):
Some comments

src/main/java/nl/tudelft/cse1110/andy/execution/process/ExternalProcess.java
```diff
@@ -0,0 +1,79 @@
...
+        // In case await was called but no process was launched.
+        if (process == null) {
+            return;
+        }
+
+        while (process.isAlive() && !output.toString().contains(endSignal)) { }
+    }
+
+    public void kill() {
+        if (process == null || !process.isAlive()) {
```
mauricioaniche, line 55: maybe extract this to a boolean variable that explains what `(process == null || !process.isAlive())` mean

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunExternalProcessStep.java
martinmladenov: This blocks the thread until either the process exits on its own or a predefined string is read from the output (`endSignal`).

mauricioaniche: Oh then this is wrong! Imagine that we are going to run a web app here. This web app must be running when we run the tests. Do we really need this await?

martinmladenov: This blocks the execution until we read a string from stdout - for instance, we could use this to wait until the web app initialises before running the tests.

martinmladenov: I'll rename `endSignal` to `initSignal` as it makes more sense that way.

src/main/java/nl/tudelft/cse1110/andy/execution/process/ExternalProcess.java
martinmladenov: I moved the run method to an inner class since it's only used for reading output. The ExternalProcess class no longer has to extend `Thread` (or implement `Runnable`). The inner class implements `Runnable`.

src/main/java/nl/tudelft/cse1110/andy/config/RunConfiguration.java
```diff
@@ -51,4 +52,8 @@ public Mode mode() {
         return Mode.PRACTICE;
     }
 
+    public ExternalProcess externalProcess() {
+        return null;
```
mauricioaniche, line 56: One idea to avoid returning null here, and later having to check about nulls everywhere, is to simply create a `NoExternalProcess` class that implements this `ExternalProcess` interface. 

This class doesn't do anything. Simply pretends that the external process was executed. See? That would avoid having to do ifs. I implemented something somewhat similar in the Writer. See the `EmptyWriter` class.

This is called the Null Object Pattern: https://sourcemaking.com/design_patterns/null_object

mauricioaniche (commented):
This PR is very good! I have only minor code quality comments.

I understand you will still write one more test to simulate a webapp, right?

src/main/java/nl/tudelft/cse1110/andy/result/ResultBuilder.java
```diff
@@ -253,6 +253,14 @@ public Result build() {
             GradeWeight weights = GradeWeight.fromConfig(ctx.getRunConfiguration().weights());
 
             int finalGrade = calculateFinalGrade(grades, weights);
+
+            if (ctx.getExternalProcess() != null) {
+                String errMsg = ctx.getExternalProcess().getErr();
+                if (errMsg != null) {
```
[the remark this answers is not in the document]
martinmladenov: This can be interpreted in 2 ways - the process exited with an exit code different than 0, or printed errors to stderr. Do we care about messages in stderr if the process exits with exit code 0 regardless?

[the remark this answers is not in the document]
martinmladenov: I made it so that error messages are ignored if they don't cause the process to crash. Some programs print warnings to stderr and we might want to ignore those.

mauricioaniche (approved):
Beautiful PR! If you are done, just merge it!

PR #48, opened 2021-09-18
Fix bug jacoco nested classes

mauricioaniche:
Jacoco was crashing in case inner classes were there. This PR basically fixes the name of the .class file.


mauricioaniche:
@martinmladenov please, check the assertions. Are those the values we are indeed expecting? Basically full coverage?

martinmladenov (approved):
I ran jacoco manually and the values are correct.

PR #49, opened 2021-09-26
Add version information

martinmladenov:
Added the Maven version, latest git commit and build timestamp in the output generated by StandardResultWriter. Added tests for the new behaviour.

This changed the parameters accepted by StandardResultWriter, so a modification in the Mojo is also necessary - https://github.com/cse1110/andy-maven-plugin/pull/4

### DIFF
--- a/src/main/java/nl/tudelft/cse1110/andy/AndyOnWebLab.java
+++ b/src/main/java/nl/tudelft/cse1110/andy/AndyOnWebLab.java
@@ -1,6 +1,7 @@
 package nl.tudelft.cse1110.andy;
 
 import nl.tudelft.cse1110.andy.execution.mode.Action;
+import nl.tudelft.cse1110.andy.utils.PropertyUtils;
 import nl.tudelft.cse1110.andy.writer.standard.RandomAsciiArtGenerator;
 import nl.tudelft.cse1110.andy.writer.weblab.WebLabResultWriter;
 
@@ -23,7 +24,7 @@ public class AndyOnWebLab {
         if (workDir == null) { System.out.println("No WORKING_DIR environment variable."); System.exit(-1); }
         if (outputDir == null) { System.out.println("No OUTPUT_DIR environment variable.");  System.exit(-1); }
 
-        WebLabResultWriter writer = new WebLabResultWriter(new RandomAsciiArtGenerator());
+        WebLabResultWriter writer = new WebLabResultWriter(PropertyUtils.getVersionInformation(), new RandomAsciiArtGenerator());
         new Andy(getAction(action), workDir, outputDir, writer).run();
     }
 
--- a/src/main/java/nl/tudelft/cse1110/andy/AndyOnWebLab.java
+++ b/src/main/java/nl/tudelft/cse1110/andy/AndyOnWebLab.java
@@ -24,7 +24,7 @@ public class AndyOnWebLab {
         if (workDir == null) { System.out.println("No WORKING_DIR environment variable."); System.exit(-1); }
         if (outputDir == null) { System.out.println("No OUTPUT_DIR environment variable.");  System.exit(-1); }
 
-        WebLabResultWriter writer = new WebLabResultWriter(PropertyUtils.getVersionInformation(), new RandomAsciiArtGenerator());
+        WebLabResultWriter writer = new WebLabResultWriter();
         new Andy(getAction(action), workDir, outputDir, writer).run();
     }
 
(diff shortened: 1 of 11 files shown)


mauricioaniche (approved):
LGTM!

If ready, merge it!

mauricioaniche:
PS: one way to avoid changing the constructor is to overload it, and have one version that simply makes the static call directly. It then avoids the clients to have to handle internal Andy stuff. Maybe that's even better.

src/main/java/nl/tudelft/cse1110/andy/AndyOnWebLab.java
```diff
@@ -23,7 +24,7 @@ public static void main(String[] args) {
         if (workDir == null) { System.out.println("No WORKING_DIR environment variable."); System.exit(-1); }
         if (outputDir == null) { System.out.println("No OUTPUT_DIR environment variable.");  System.exit(-1); }
 
-        WebLabResultWriter writer = new WebLabResultWriter(new RandomAsciiArtGenerator());
+        WebLabResultWriter writer = new WebLabResultWriter(PropertyUtils.getVersionInformation(), new RandomAsciiArtGenerator());
```
mauricioaniche, line 27: In here, let's use the constructor that only requires the ascii generator. (This constructor calls the property version).

In general words, no client classes should need to know about this property version thingy.

src/main/java/nl/tudelft/cse1110/andy/writer/standard/StandardResultWriter.java
```diff
@@ -19,17 +20,23 @@
...
     private StringBuilder toDisplay = new StringBuilder();
     private List<Highlight> highlights = new ArrayList<>();
 
-    public StandardResultWriter(RandomAsciiArtGenerator asciiArtGenerator) {
+    public StandardResultWriter(VersionInformation versionInformation, RandomAsciiArtGenerator asciiArtGenerator) {
+        this.versionInformation = versionInformation;
         this.asciiArtGenerator = asciiArtGenerator;
     }
 
+    public StandardResultWriter(VersionInformation versionInformation) {
```
mauricioaniche, line 34: Do we really need this one?

src/main/java/nl/tudelft/cse1110/andy/writer/weblab/WebLabResultWriter.java
```diff
@@ -15,8 +16,8 @@
 
 public class WebLabResultWriter extends StandardResultWriter {
 
-    public WebLabResultWriter(RandomAsciiArtGenerator asciiArtGenerator) {
-        super(asciiArtGenerator);
+    public WebLabResultWriter(VersionInformation versionInformation, RandomAsciiArtGenerator asciiArtGenerator) {
```
mauricioaniche, line 19: Unless needed for tests, I would not create this constructor

mauricioaniche: Ok, it's needed for the tests. So I'd create an extra constructor that only receives ascii generator, and that's the one clients will use.

Maybe even make this one that receives the property to be with default visibility, so that only the test can use it

mauricioaniche (changes requested):
Client classes should not need to pass the property version. Having a constructor that receives it for testing purposes is fine, but this should be abstracted away from client classes.

I hope the in-line comments help you in understanding precisely what I mean!

src/main/java/nl/tudelft/cse1110/andy/writer/weblab/WebLabResultWriter.java
martinmladenov: I don't think we can do that - the test is in a different package.

src/main/java/nl/tudelft/cse1110/andy/AndyOnWebLab.java
martinmladenov: I made it so that it uses an empty constructor. Is there any point in passing the ascii generator here, it's also an internal component?

mauricioaniche: My rule of thumb is: if the dependency is a sort of a tool (i.e., something with some logic in it), like the case of the `RandomAsciiArtGenerator`, I pass it via the constructor, and I do not mind requiring clients to instantiate it and pass it over. This reminds clients of the dependencies the class has.

Now, the version information is just "some information"; it's not a component per se. We will never replace it by something else. If we replaced it by simple static method calls, and no dependencies, we'd be happy too. We'd require the test to have the same .properties file in the resources folder, that's not a big deal. Passing it as a dependency enables us to mock, which is also good and clean, but then, I'd make sure clients are not required to pass it when instantiating the main class.

Reading suggestion: Growing Object-Oriented Systems Guided by Tests, chapter 6!

src/main/java/nl/tudelft/cse1110/andy/writer/weblab/WebLabResultWriter.java
mauricioaniche: True, we can leave it public.

mauricioaniche (approved):
LGTM!

If ready, just merge!

PR #50, opened 2021-10-02
Print explanation when test score is 0/100

martinmladenov:
Print a suitable explanation to the output when the weblab score is overridden to 0/100 due to the selected mode or action

### DIFF
--- a/src/main/java/nl/tudelft/cse1110/andy/writer/standard/StandardResultWriter.java
+++ b/src/main/java/nl/tudelft/cse1110/andy/writer/standard/StandardResultWriter.java
@@ -1,6 +1,8 @@
 package nl.tudelft.cse1110.andy.writer.standard;
 
 import nl.tudelft.cse1110.andy.execution.Context;
+import nl.tudelft.cse1110.andy.execution.mode.Action;
+import nl.tudelft.cse1110.andy.execution.mode.Mode;
 import nl.tudelft.cse1110.andy.execution.mode.ModeActionSelector;
 import nl.tudelft.cse1110.andy.result.*;
 import nl.tudelft.cse1110.andy.utils.ExceptionUtils;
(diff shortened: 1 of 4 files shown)


src/main/java/nl/tudelft/cse1110/andy/writer/standard/StandardResultWriter.java
```diff
@@ -106,6 +110,37 @@ private void printFinalGrade(Context ctx, Result result) {
...
+
+        if (mas.getMode() == Mode.EXAM) {
+            reason.append("shown as 0/100 during the exam and will be calculated after the exam is over.");
+        } else {
+            reason.append("0/100 ");
+            if (mas.getAction() != Action.FULL_WITH_HINTS && mas.getAction() != Action.FULL_WITHOUT_HINTS) {
+                reason.append("as you are only ");
+                if (mas.getAction() == Action.TESTS)
+                    reason.append("checking if your tests pass");
+                else if (mas.getAction() == Action.META_TEST)
```
mauricioaniche, line 131: This should never happen. This is an internal mode used only when we are running the meta tests. What if the method that gets the mode throws an exception in case we get this one? As a sort of post condition.

mauricioaniche (approved):
I like the message. I have only one minor comment. You can merge once you are ready.

src/main/java/nl/tudelft/cse1110/andy/writer/standard/StandardResultWriter.java
martinmladenov: Which method should throw an exception? This one or `getAction()` in `ModeActionSelector`?

martinmladenov: Done!

mauricioaniche (approved):
LGTM!

PR #54, opened 2021-10-17
Move generic failure handling from ResultBuilder to StandardResultWriter

martinmladenov:


src/main/java/nl/tudelft/cse1110/andy/result/ResultBuilder.java
```diff
@@ -259,15 +259,34 @@ public Result build() {
...
+            this.checkExternalProcessExit();
+
             int finalGrade = calculateFinalGrade(grades, weights);
 
-            ExternalProcess process = ctx.getExternalProcess();
-            if (!process.hasExitedNormally()) {
-                genericFailureMessage = String.format("External process crashed with exit code %d: %s",
-                        process.getExitCode(), process.getErrorMessages());
+            GenericFailure genericFailure = null;
+
```
mauricioaniche, line 267: Can we simply call a "buildAnyPossibleFailure()"? This method would then create a GenericFailure if one needed to be created. This method would also call the checkExternalProcess thingy

src/main/java/nl/tudelft/cse1110/andy/result/ResultBuilder.java
```diff
@@ -292,8 +311,15 @@ private int calculateFinalGrade(GradeValues grades, GradeWeight weights) {
     public boolean hasFailed() {
         boolean compilationFailed = compilation!=null && !compilation.successful();
         boolean unitTestsFailed = testResults != null && testResults.didNotGoWell();
-        boolean genericFailureHappened = genericFailureMessage != null;
+        boolean hasGenericFailure = hasGenericFailure();
+
+        return compilationFailed || unitTestsFailed || hasGenericFailure;
+    }
+
+    private boolean hasGenericFailure() {
```
mauricioaniche, line 319: Also thinking of a way to move this method to the new GenericFailure class

mauricioaniche (commented):
Thanks for the PR!

A few comments

src/main/java/nl/tudelft/cse1110/andy/result/ResultBuilder.java
martinmladenov: Thank you for the suggestion, I extracted a method that builds the generic failure using the class variables. However, there is currently no way to move the call to `checkExternalProcessExit` to this method, as `checkExternalProcessExit` is used to set a failure flag if the process has crashed, which is then used in `calculateFinalGrade`, so it has to be called before calculating the grade.

```java
this.checkExternalProcessExit();

int finalGrade = calculateFinalGrade(grades, weights);

GenericFailure genericFailure = this.buildGenericFailure();
```

martinmladenov: I refactored the way GenericFailure objects are built and stored within the ResultBuilder class, and now it is possible to move this method call.

src/main/java/nl/tudelft/cse1110/andy/result/ResultBuilder.java
martinmladenov: I made `GenericFailure.build()` return null if the arguments provided to it mean that there is no generic failure (i.e. they are all null). This makes the `hasGenericFailure` method obsolete, so I removed it entirely.

src/main/java/nl/tudelft/cse1110/andy/result/GenericFailure.java
```diff
@@ -0,0 +1,59 @@
...
+        this.externalProcessExitCode = externalProcessExitCode;
+        this.externalProcessErrorMessages = externalProcessErrorMessages;
+    }
+
+    public static GenericFailure build(String genericFailure, String stepName, String exceptionMessage, Integer externalProcessExitCode, String externalProcessErrorMessages) {
+        boolean genericFailureHappened = genericFailure != null || stepName != null || exceptionMessage != null;
+        boolean externalProcessFailed = externalProcessExitCode != null || externalProcessErrorMessages != null;
+
+        if (!genericFailureHappened && !externalProcessFailed) {
+            return null;
```
mauricioaniche, line 23: Should we return null or should we return a "NoFailure" object to avoid returning null?

src/main/java/nl/tudelft/cse1110/andy/result/ResultBuilder.java
```diff
@@ -292,8 +313,8 @@ private int calculateFinalGrade(GradeValues grades, GradeWeight weights) {
     public boolean hasFailed() {
         boolean compilationFailed = compilation!=null && !compilation.successful();
         boolean unitTestsFailed = testResults != null && testResults.didNotGoWell();
-        boolean genericFailureHappened = genericFailureMessage != null;
+        boolean hasGenericFailure = genericFailureObject != null;
```
mauricioaniche, line 316: Instead of comparing to null, this could be a method call to genericFailure.someMethod().

src/main/java/nl/tudelft/cse1110/andy/result/GenericFailure.java
mauricioaniche: Or return a Optional ? At least it becomes explicit that this can be null.

mauricioaniche (approved):
LGTM!

Suggestion: either use an Optional or create a class, as to avoid retuning null and comparing to null.

src/main/java/nl/tudelft/cse1110/andy/result/GenericFailure.java
martinmladenov: I made it so that a GenericFailure object is always created (it is never null in a valid result), and added a `hasFailure` method in GenericFailure, which checks if any of its internal fields are non-null.

martinmladenov:
@mauricioaniche 
I made it so that the GenericFailure is never null in Result. Instead, an empty GenericFailure with all of its internal failure types set to null is created when the ResultBuilder is generated, and it is rebuilt if an error occurs.

Furthermore, I replaced the return values of all getters in GenericFailure with Optionals in order to emphasise that those values are often not present. What do you think?

mauricioaniche (approved):
LGTM! If it's ready, merge it!!

PR #62, opened 2021-10-28
Add `runWithoutSecurity` method

### LINKED ISSUE #61: Create `run` method without security manager
Whenever running `andy` locally, there is no need for the `AndySecurityManager` to take over control of the permissions, and thus this should be disabled when running locally. For this, a method `runWithoutSecurityManager` should be created which does not add the `AddSecurityManagerStep` to the execution flow, this way no security manager is added to the `andy` process.

thijsnulle:
This PR adds a `runWithoutSecurityManager` method which will be used in the `andy-maven-plugin` to ensure that whenever Andy is run locally, the security manager won't interfere with anything since it isn't necessary to check for students breaking Andy.

Closes #61


src/main/java/nl/tudelft/cse1110/andy/execution/ExecutionFlow.java
```diff
@@ -50,13 +50,12 @@ public void addSteps(List<ExecutionStep> steps) {
...
-        return new ExecutionFlow(ctx, result, writer);
+    public static ExecutionFlow build(Context ctx, ResultBuilder result, ResultWriter writer, boolean addSecurityManager) {
+        return new ExecutionFlow(ctx, result, writer, addSecurityManager);
     }
 
     private List<ExecutionStep> basicSteps() {
         return Arrays.asList(
-                new SetSecurityManagerStep(),
                 new OrganizeSourceCodeStep(),
                 new SourceCodeSecurityCheckStep(),
```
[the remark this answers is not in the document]
thijsnulle: Updated it to add both if the boolean `addSecurityManager` is true. Do you think renaming it something along the lines of `addSecurity` or `enableSecurity` would make more sense?

src/main/java/nl/tudelft/cse1110/andy/execution/ExecutionFlow.java
```diff
@@ -15,9 +15,14 @@
     private LinkedList<ExecutionStep> steps;
     private final ResultWriter writer;
 
-    private ExecutionFlow(Context ctx, ResultBuilder result, ResultWriter writer) {
+    private ExecutionFlow(Context ctx, ResultBuilder result, ResultWriter writer, boolean addSecurityManager) {
```
mauricioaniche, line 18: Can we have two methods here, buildWithSecurityChecks and build ?

Internally, we can use a Boolean to differentiate and avoid duplicate code, but for the clients of the API, i would avoid Boolean flags like this one.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunMetaTestsStep.java
```diff
@@ -120,7 +120,7 @@ private ResultBuilder runMetaTest(Context ctx, DirectoryConfiguration dirCfg, Fi
 
         ResultBuilder metaResult = new ResultBuilder(metaCtx, new GradeCalculator());
 
-        ExecutionFlow flow = ExecutionFlow.build(metaCtx, metaResult, new EmptyWriter());
+        ExecutionFlow flow = ExecutionFlow.build(metaCtx, metaResult, new EmptyWriter(), true);
```
mauricioaniche, line 123: Oh here, I guess the security check should also be false in case we are running it without security checks, right?

mauricioaniche (commented):
Two comments

src/main/java/nl/tudelft/cse1110/andy/execution/ExecutionFlow.java
mauricioaniche: If that's too much work given the MetaStep, then it's ok for now.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunMetaTestsStep.java
mauricioaniche: Maybe the info about whether it's running in secure mode should go to the Context class, so that you know which mode to run in the meta test.

src/main/java/nl/tudelft/cse1110/andy/execution/ExecutionFlow.java
thijsnulle: Updated this to have two methods, `build` and `buildWithoutSecurityChecks`. I do not understand what is extra about the `MetaStep` but that might just be my misunderstanding of how the meta step works

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunMetaTestsStep.java
thijsnulle: I think running it with security checks is a stronger assertion, and I think we want our tests working with the security manager. Also, I moved the boolean value to the `Context` class since this made indeed more sense.

mauricioaniche: If we run with the security mode off, the meta steps should also run _without_ the security steps. (With this current version of the code, even if Andy is run without security checks, it will do the security checks for the meta tests.)

What's happening here is that we replace the code with the mutated version, and run "Andy on top of Andy". That's why you see a new _ExecutionFlow_ being instantiated.

thijsnulle: I understand it now, I just assumed this was only run in our local tests but I get it. I changed it and built the `ExecutionFlow` based on the boolean value of the `Context` passed to the method.

martinmladenov (commented):
Looks good. I have just one comment: in the method and field names, there are many references to "security manager". I think it would be better to change that to just "security" (i.e. `runWithoutSecurity`) or "security features", or something else that's more general than "security manager" since that also disables other security checks we perform (like the `SourceCodeSecurityCheckStep`, which has nothing to do with the SecurityManager, as well as other security steps we might add in the future).

martinmladenov:
We should also modify the Maven plugin to run Andy without security features by default, since the security manager sometimes fails for no obvious reason depending on OS, JVM version, etc, and this could cause issues for students who want to run Andy locally. There should also be an optional argument to run it with security checks so that we can execute the security steps when we run the integration tests with the assignments repo.

martinmladenov (approved):

thijsnulle:
@martinmladenov 
> We should also modify the Maven plugin to run Andy without security features by default, since the security manager sometimes fails for no obvious reason depending on OS, JVM version, etc, and this could cause issues for students who want to run Andy locally. There should also be an optional argument to run it with security checks so that we can execute the security steps when we run the integration tests with the assignments repo.

Yes, this is something that we will change, but for this implementation, the method needs to be added to Andy itself. Once this PR is merged, it will be updated in the andy-maven-plugin repository!

thijsnulle:
@martinmladenov 
> Looks good. I have just one comment: in the method and field names, there are many references to "security manager". I think it would be better to change that to just "security" (i.e. `runWithoutSecurity`) or "security features", or something else that's more general than "security manager" since that also disables other security checks we perform (like the `SourceCodeSecurityCheckStep`, which has nothing to do with the SecurityManager, as well as other security steps we might add in the future).

I am with you on this one, do you think the method name `runWithoutSecurity` and the variable name `enableSecurity` in the `Context` class are good?

martinmladenov:
> @martinmladenov
> 
> > Looks good. I have just one comment: in the method and field names, there are many references to "security manager". I think it would be better to change that to just "security" (i.e. `runWithoutSecurity`) or "security features", or something else that's more general than "security manager" since that also disables other security checks we perform (like the `SourceCodeSecurityCheckStep`, which has nothing to do with the SecurityManager, as well as other security steps we might add in the future).
> 
> I am with you on this one, do you think the method name `runWithoutSecurity` and the variable name `enableSecurity` in the `Context` class are good?

Yes, this sounds good to me!

martinmladenov:
> @martinmladenov
> 
> > Looks good. I have just one comment: in the method and field names, there are many references to "security manager". I think it would be better to change that to just "security" (i.e. `runWithoutSecurity`) or "security features", or something else that's more general than "security manager" since that also disables other security checks we perform (like the `SourceCodeSecurityCheckStep`, which has nothing to do with the SecurityManager, as well as other security steps we might add in the future).
> 
> I am with you on this one, do you think the method name `runWithoutSecurity` and the variable name `enableSecurity` in the `Context` class are good?

@thijsnulle 
Yes, this sounds good to me!

Arraying:
This still seems quite relevant to me. I can sit down and fix the merge conflicts, that's no problem. Shall we go ahead with a merge here guys?

mauricioaniche:
The goal here was to make sure that our maven plugin doesn't use the security manager. After all, it is meant for people to run it locally. It's not a big thing, though, if we run it with the security manager now that it is working properly. I'll leave it up to you and Martin to decide that!

martinmladenov:
The security manager has to be disabled when running via Maven as its behaviour can be inconsistent on different JVMs (this is a Java issue, it's not something we can fix). I'll resolve the merge conflicts since they seem to affect mainly code that I have written.

src/main/java/nl/tudelft/cse1110/andy/execution/metatest/library/LibraryMetaTest.java
```diff
@@ -99,6 +99,10 @@ private ResultBuilder runMetaTest(Context ctx, DirectoryConfiguration dirCfg, Fi
         metaCtx.setDirectoryConfiguration(metaDirCfg);
         metaCtx.setLibrariesToBeIncluded(ctx.getLibrariesToBeIncluded());
 
+        if(!ctx.isSecurityEnabled()){
```
Arraying, line 102: Are these meta tests run when running Andy through Maven?

martinmladenov: Yes, they are. Running it through Maven runs the same steps as running it normally.

 Meta tests are run in their own "sub-context" and we want to configure this context not to use security if the main context doesn't.

Arraying: LGTM then!

Arraying:
Let's merge this, if there's any issues we can fix them as we find them!

PR #63, opened 2021-10-30
Implement success message

martinmladenov:


src/main/java/nl/tudelft/cse1110/andy/config/RunConfiguration.java
```diff
@@ -57,4 +57,8 @@ public ExternalProcess externalProcess() {
         return new EmptyExternalProcess();
     }
 
+    public String successMessage() {
+        return null;
+    }
```
martinmladenov, line 62: The only thing I'm not sure is this: should this be a method that returns a `String` (or `null` if nonexistent), or would it be better if it is an `Optional<String>`? I feel like an `Optional` would make configurations seem a bit "cluttered".

mauricioaniche (approved):
LGTM!

I'd keep it as you did, returning null.

PR #65, opened 2021-11-09
Remove "getClass" from security keyword blacklist as it can cause false positives

martinmladenov:
`getClass()` is sometimes used in the `equals` methods of classes in order to compare their type. It can cause false positives, so I think it's best to remove it from the blacklist.


mauricioaniche:
Why? I'd keep this method. And remove any exercise that requires reflection. It's safer I'd say. (Reflection is not something fun to have in an exercise anyways)

martinmladenov:
@mauricioaniche 
> Why? I'd keep this method. And remove any exercise that requires reflection. It's safer I'd say. (Reflection is not something fun to have in an exercise anyways)

The problem is that `getClass` specifically can be used in applications other than reflection. For example:
```java
@Override
public boolean equals(Object o) {
    if (this == o) return true;

    if (o == null || getClass() != o.getClass()) return false;

    OwnerInfo ownerInfo = (OwnerInfo) o;
    return Objects.equals(name, ownerInfo.name) &&
           Objects.equals(address, ownerInfo.address);
}
```

mauricioaniche:
That makes sense!

mauricioaniche (approved):

PR #66, opened 2021-11-16
Improve WebLab selenium tests

martinmladenov:
Added tests for multiple action buttons and a workflow that runs the Selenium tests and checks if the Andy version deployed on WebLab works as expected every day at 06:00 UTC.


src/test/java/selenium/WebLabSeleniumTestBase.java
```diff
@@ -0,0 +1,79 @@
...
+                 "course containing the test assignments.");
+            return;
+        }
+
+        this.weblabUsername = weblabCredentials[0];
+        this.weblabPassword = weblabCredentials[1];
+
+        FirefoxOptions options = new FirefoxOptions();
+        if (StringUtils.isNotEmpty(System.getenv(WEBLAB_SELENIUM_HEADLESS_ENV_VAR))) {
+            options.setHeadless(true);
```
mauricioaniche, line 55: I didn't know you could run Firefox headless! Super cool!

src/test/java/selenium/ExamAssignmentTests.java
```diff
@@ -0,0 +1,106 @@
...
+                .has(not(metaTestsPassing(2)))
+                .has(not(metaTests(3)))
+                .has(not(metaTestPassing("does not update")))
+                .has(not(metaTestFailing("does not add points if post is not featured")))
+                .has(noFinalGrade())
+                .has(not(fullGradeDescription("Branch coverage", 2, 2, 0.25)))
+                .has(not(fullGradeDescription("Mutation coverage", 3, 4, 0.25)))
+                .has(not(fullGradeDescription("Code checks", 17, 18, 0.25)))
+                .has(not(fullGradeDescription("Meta tests", 2, 3, 0.25)))
+                // .has(zeroScoreExplanation()) TODO
```
mauricioaniche, line 50: Delete this line?

mauricioaniche (approved):
LGTM!

One remark: given that we are hard coding the assignment number in the test, what about a code comment that explains what this assignment is and the url so that one can easily copy and paste into their browser if they want to see it?

martinmladenov:
Should I make the assignment id and the user id environment variables like the username and the password, or is it better to hardcode them?

mauricioaniche:
I prefer it in the code, much easier to change if we want. This is also a real part of the test scenario. So, one more reason to keep it there!

src/test/java/selenium/ExamAssignmentTests.java
martinmladenov: I had previously deactivated this since the version deployed on WebLab did not support this feature yet. I activated it again since the current version does support this.

martinmladenov:
I made some changes to the code. Namely, I used `WebDriverWait` to fix an issue that was introduced at some point (most likely because of a WebLab editor code change). This issue made our tests flaky since the editor does not render immediately in some cases, and Selenium could not find it while trying to submit the solution, which made the tests randomly fail. To fix this, I made Selenium wait until the editor has rendered before continuing with the execution.

mauricioaniche (approved):
LGTM

PR #67, opened 2021-11-18
Selenium and PIT Fixes

Arraying:
__Roadmap__:
- [x] Explain the bug and add preventative measures (@Arraying)
- [x] Ensure Selenium works in production by patching the image (@Arraying)
- [x] Fix PIT not working with jqwik (81c24ad95f561632313c201e18e3fddcd4872f9e) (@Arraying)
- [x] Fix the security manager issues with Selenium (@martinmladenov)
- [x] Fix the issue where Selenium tests don't compile when running Andy as a jar compiled via Maven

# Selenium

## What happened

Essentially an XY problem.  Long story short, this has nothing to do with dependencies or classloaders. Selenium requires a driver to work. When this is not specified properly (or it straight up does not exist), this will throw an exception. This small exception blew up and caused some very strange reflection behaviour.

The problem was also made worse by the complexity of the test suite, which made debugging very difficult. I'm not 100% sure on the exact details but it went something like this:
- The first test tried to initialize the test suite, which caused an exception `java.lang.ExceptionInInitializerError`.
- The initialization fails, therefore somehow the class definition doesn't get registered in the JVM.
- Subsequent test suite interactions now throw a `Could not initialize class delft.PetClinicTest` due to this lack of class definition.

## The actual fix

Add a system property when running the jar that specifies the driver path:
```
-Dwebdriver.gecko.driver=/path/to/driver
```

## Moving forward

The most difficult part in debugging was figuring out where the error came from. Once I isolated it I printed out the entire stack trace. When one error is caused by another, it's not guaranteed that the topmost error message (i.e. the one that is printed to the user in the end) contains all the error messages. The issue became apparent really quickly when I printed out the entire stacktrace.

As such, I made a loop that prints all stacktraces in their entirety to the console. That way, all of the error messages will get printed and if something is behaving weirdly the developer has a much better idea than the vague Java reflection messages.

# PIT

## What happened

For some reason, the test classes were picked up but the JUnit tests could not be read. After comparing the settings and classpaths of the version IntelliJ runs vs. the version that is run by Maven, I did a set difference. This showed that for some reason, the Maven version did not include `jqwik-engine-1.5.1.jar` in its classpath. 

Inspecting the Maven dependency structure of `andy`, it looked like this:
img

The *(runtime)* explains why it was not included in the classpath. I'm not entirely sure but I think it's because the Maven plugin only includes dependencies that need to be compiled when the plugin is being run.

## The actual fix

Explicitly adding a dependency for the engine and setting its scope to `compile` overrides this *(runtime)* setting.


src/main/java/nl/tudelft/cse1110/andy/execution/ExecutionFlow.java
```diff
@@ -56,9 +56,9 @@ public static ExecutionFlow build(Context ctx, ResultBuilder result, ResultWrite
 
     private List<ExecutionStep> basicSteps() {
         return Arrays.asList(
-                new SetSecurityManagerStep(),
+                //new SetSecurityManagerStep(),
```
mauricioaniche, line 59: Should we disable it? Please, confirm, @martinmladenov

mauricioaniche (commented):
Oh, that's a good catch! For some reason, we were actually using HtmlUnit as a driver, so no need for the driver.

Shall we try it with html unit instead of Firefox (or whatever we are using)? That should simplify things a lot too.

I also have a bunch of comments in the code. Should we really disable the security mechanism?

mauricioaniche:
Ah, just saw the TODO from @martinmladenov to fix the Selenium issues with the security manager.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunJUnitTestsStep.java
```diff
@@ -48,6 +49,15 @@ public void execute(Context ctx, ResultBuilder result) {
 
             TestExecutionSummary summary = listener.getSummary();
 
+            /* Print all the exceptions that were occurred.
+               In most cases, these will be useless.
+               However this gives very valuable information with bugs like the Selenium bug.
+             */
+            for (int i = 0; i < summary.getFailures().size(); i++) {
+                console.println("\nFailure Report [" + (i + 1) + "/" + summary.getFailures().size() + "]");
```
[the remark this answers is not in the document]
martinmladenov: Maybe putting it into the generic failure message might be better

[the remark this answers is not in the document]
Arraying: BTW, Selenium or one of the drivers uses SLF4J to log. That might also need to be taken care of if you wish to clean stdout. Maybe a custom log adapter?

src/main/java/nl/tudelft/cse1110/andy/execution/ExecutionFlow.java
Arraying: Currently the plan is for him to work on this branch and to uncomment these lines when everything is working again. That way we can at least test that the tests are working without having to wait for the security manager to work. Maybe I should have specified 😅

Arraying:
@mauricioaniche Sure, let's try it headless. For this I think we need to add a driver too, but that should just be a Maven dependency. And then we don't need to worry about system properties... I hope. I have some form of solution but that's still using the Firefox driver:
```java
protected static WebDriver driver = new FirefoxDriver();
```
Shall I just replace `FirefoxDriver` with `HtmlUnitDriver` or is there any other changes that need to be made?

mauricioaniche:
I think just HtmlUnitDriver and the dependency should be enough. No need for env variables or properties!

Arraying:
`HtmlUnitDriver` works as intended without needing any drivers and/or system properties. I tested this on the pet clinic solution using `BrowserVersion.FIREFOX` and both with and without JavaScript support. In the context of the pet shop, I don't think using or not using JavaScript makes a difference.

I'll also mark point 2 as complete on the checklist - the image will work since now we do not need a driver, just a jar with the driver as a dependency.

martinmladenov:
Just a thought: We could include a custom selenium version that substitutes all drivers with `HtmlUnitDriver` (and/or makes them headless as non-headless browsers crash in docker) so that students can test their code with an actual non-headless browser locally and see what's happening and submit their code without the need to modify it first.

mauricioaniche:
That would be nice, @martinmladenov! Let's open a TODO for that.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunJUnitTestsStep.java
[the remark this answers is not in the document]
Arraying: We can forget about this when using mutation testing. Some of the mutants (rightfully) cause assertion errors that this picks up. Definitely unhelpful.

src/main/java/nl/tudelft/cse1110/andy/execution/security/AndySecurityManager.java
```diff
@@ -122,22 +127,31 @@ private boolean checkLoggingPermission(Permission perm) {
...
+                || seleniumInternal && checkSeleniumRuntimePermissions(perm)
+                || checkNormalRuntimePermissions(perm)) {
                 return true;
             }
         }
         return false;
     }
 
+    private boolean checkSeleniumRuntimePermissions(Permission perm) {
+        return perm.getName().equals("modifyThread");
```
mauricioaniche, line 143: A comment explaining what "modifythread" is may help future understanding

mauricioaniche (approved):
LGTM, with a small inline comment.

I suppose having 0 in the weight of coverage and mutation and so makes the step to not run. Am I right ?

martinmladenov:
@mauricioaniche 
> I suppose having 0 in the weight of coverage and mutation and so makes the step to not run. Am I right ?

Weights don't affect whether a step runs. However, I skip the mutation and coverage steps by running Andy with the `TESTS` action:
```java
Result result = run(Action.TESTS, "EmptyLibrary", "SeleniumOnePassingOneFailingSolution", "SeleniumSimpleWebpageConfiguration");
```

src/main/java/nl/tudelft/cse1110/andy/execution/security/AndySecurityManager.java
martinmladenov: I'm not sure what exactly this is needed for since there's no documentation on the permissions Selenium requires. I added a comment simply explaining that it's necessary for Selenium, is that sufficient?

mauricioaniche: Maybe what modifyThread means. Is this a method in selenium or a Java thing ?

martinmladenov: It's a Java thing, it's a qualifier of `RuntimePermission`. It allows the process to execute methods such as `interrupt()`, `suspend()`, `resume()` on threads. However, I am not sure why exactly Selenium requires that as it is not specified anywhere. https://docs.oracle.com/en/java/javase/15/security/permissions-jdk1.html

mauricioaniche: Maybe that's the comment then :)

martinmladenov (commented):
Maybe it would be nice to rename `SourceCodeSecurityCheckStep` to something like `SourceCodeKeywordCheckStep` since it's no longer used just for security

src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
```diff
@@ -64,7 +66,14 @@ private boolean checkForKeywords(String code, ResultBuilder result) {
...
                 "setAccessible", reflectionMsg
-        );
+        ));
+        keywords.put("ChromeDriver", seleniumUnsupportedDriverMsg);
+        keywords.put("ChromiumDriver", seleniumUnsupportedDriverMsg);
+        keywords.put("EdgeDriver", seleniumUnsupportedDriverMsg);
+        keywords.put("FirefoxDriver", seleniumUnsupportedDriverMsg);
+        keywords.put("InternetExplorerDriver", seleniumUnsupportedDriverMsg);
+        keywords.put("OperaDriver", seleniumUnsupportedDriverMsg);
+        keywords.put("SafariDriver", seleniumUnsupportedDriverMsg);
```
martinmladenov, line 76: Why didn't you put those new keywords inside `Map.of`?

Arraying: Line 59:
```java
// An immutable map only supports 10 hardcoded KVs, we need more, hence HashMap wrapper.
```
The constructor only goes up to 10 keys and values and with these we have more.

Arraying:
> Maybe it would be nice to rename `SourceCodeSecurityCheckStep` to something like `SourceCodeKeywordCheckStep` since it's no longer used just for security

Sure, let's do that.

src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
```diff
@@ -55,7 +55,9 @@ private boolean checkPackageName(String code, ResultBuilder result) {
 
     private boolean checkForKeywords(String code, ResultBuilder result) {
         String reflectionMsg = "Using reflection in your code is not allowed";
-        var keywords = Map.of(
+        String seleniumUnsupportedDriverMsg = "Please use HtmlUnitDriver as WebLab does not have a browser installed";
```
mauricioaniche, line 58: Instead of "WebLab", shall we say that "Andy" doesn't support it? This way the message will look nice even in the command line version.

mauricioaniche (approved):
LGTM!

I have one small inline remark. You can merge it afterwards!

src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
martinmladenov: We should change this so that it has the same format as the current keyword list. (`Map.ofEntries`)

Arraying:
I've fixed the merge conflicts, however there are still some failing tests:
- SecurityTest: RuntimeExec [fails](https://paste.gg/p/anonymous/97a0983d111043748ea195ded7600dab), I did not touch anything to do with `Runtime`
- GenericFailureTest#genericFailureHasPrecedence [fails](https://paste.gg/p/anonymous/6446758174a247faaa8c094c3b2b9e13)
- WebLabTest#allFilesGenerated [fails](https://paste.gg/p/anonymous/fbcd12546b3e46deb48c0da2259cfd5b)

Are any of these normal/known? Perhaps someone could re-run them locally and see if anything is different for them.

mauricioaniche:
No tests are expected to fail. At least not before we started this :)

martinmladenov:
Not sure why those would fail - [according to our CI](https://github.com/cse1110/andy/runs/4992851962?check_suite_focus=true), all tests pass. I'll rerun them locally just to be sure.

src/test/java/integration/GenericFailureTest.java
```diff
@@ -43,7 +57,7 @@ void genericFailureHasPrecedence() {
         assertThat(result.getGenericFailure().getExceptionMessage())
                 .isPresent()
                 .hasValueSatisfying(
-                        startsWithString("java.lang.RuntimeException: Some super error here\n" +
+                        startsWithString("java.lang.RuntimeException: Some super error here" + crlf +
```
martinmladenov, line 60: You could replace all of those changes with `System.lineSeparator()` - this returns `\r\n` on windows and `\n` on everything else.

https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#lineSeparator--

Arraying: You're right. That is infinitely times better.

Arraying:
Folks, shall we merge this sometime soon?

martinmladenov:
I'll do some final tests with docker to make sure everything still works and then we can merge!

martinmladenov (approved):
Everything works! Let's merge!

Arraying:
LGTM!

PR #69, opened 2021-11-23
Remove student name

yoonhwanjeong:
WebLab will not provide student information except anonymized id for aggregate analytics.


mauricioaniche (approved):
LGTM!

PR #70, opened 2021-11-23
Read meta data from environment

yoonhwanjeong:
Read meta data from environment according to [this](https://3.basecamp.com/4338320/buckets/22606949/todos/3837601280#__recording_4361054885).
I do not know how to run Andy in WebLab so I would like someone else to check that the output has correct meta data.

### DIFF
--- a/src/main/java/nl/tudelft/cse1110/andy/writer/weblab/WebLabResultWriter.java
+++ b/src/main/java/nl/tudelft/cse1110/andy/writer/weblab/WebLabResultWriter.java
@@ -99,7 +99,7 @@ public class WebLabResultWriter extends StandardResultWriter {
 
         Submission submission = new Submission(
                 ctx.getAction(),
-                new SubmissionMetaData("course", "studentid", "exercise"),
+                ctx.getMetaData(),
                 result
         );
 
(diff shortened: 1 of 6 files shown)


mauricioaniche:
I would like to avoid the system call inside that class. I'd suggest creating a class to represent the student data and pass it to the Context. Check the AndyOnWebLab class, that's the main method that Weblab calls!

src/main/java/nl/tudelft/cse1110/andy/writer/weblab/WebLabResultWriter.java
```diff
@@ -98,7 +98,8 @@ private void writeAnalyticsFile(Context ctx, Result result) {
             return;
 
         Submission submission = new Submission(
-                new SubmissionMetaData(System.getenv("WL_COURSE"), System.getenv("WL_STUDENT"), System.getenv("WL_ASSIGNMENT_TITLE"), ctx.getAction()),
+                ctx.getAction(),
+                ctx.getMetaData(),
```
mauricioaniche, line 102: I'd just rename the method to `getSubmissionMetaData` to make it more explicit!

mauricioaniche (approved):
One small comment only, but LGTM! You can merge after the suggestion!

PR #71, opened 2021-11-23
PIT not working with jqwik fix

Arraying:
**Note:** This relies on the changes made in the [Selenium fix](https://github.com/cse1110/andy/tree/fix/selenium) branch as the security manager goes ham with external assertions. Also, the print functionality of that branch was quite helpful.

**The actual fix:** Is contained in this commit 81c24ad95f561632313c201e18e3fddcd4872f9e, if applicable it can be cherry picked.

# What happened

For some reason, the test classes were picked up but the JUnit tests could not be read. After comparing the settings and classpaths of the version IntelliJ runs vs. the version that is run by Maven, I did a set difference. This showed that for some reason, the Maven version did not include `jqwik-engine-1.5.1.jar` in its classpath. 

Inspecting the Maven dependency structure of `andy`, it looked like this:
img

The *(runtime)* explains why it was not included in the classpath. I'm not entirely sure but I think it's because the Maven plugin only includes dependencies that need to be compiled when the plugin is being run.

# The actual fix

Explicitly adding a dependency for the engine and setting its scope to `compile` overrides this *(runtime)* setting.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,13 @@
             <version>1.5.1</version>
         </dependency>
 
+        <dependency>
+            <groupId>net.jqwik</groupId>
+            <artifactId>jqwik-engine</artifactId>
+            <version>1.5.1</version>
+            <scope>compile</scope>
+        </dependency>
+
         <!-- https://mvnrepository.com/artifact/org.mockito/mockito-core -->
         <dependency>
             <groupId>org.mockito</groupId>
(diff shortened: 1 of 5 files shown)


mauricioaniche:
This is brilliant, Paul. Thanks!

@martinmladenov once the selenium one is fixed, let's see if this branch passes the tests!!

Arraying:
This change has been adopted in [this PR](https://github.com/cse1110/andy/pull/67) in an attempt to make all tests pass. Shall we close this PR in favour of the other one? As this one is now a small subset of the Selenium PR.

mauricioaniche:
Sure! Feel free to close it!

PR #72, opened 2021-11-26
Fix selenium test flakiness

martinmladenov:


mauricioaniche (approved):
LGTM

"Status done" seems the right thing to wait for!

src/test/java/selenium/pageobjects/BasePageObject.java
```diff
@@ -0,0 +1,43 @@
...
+        wait.until(new ExpectedCondition<Boolean>() {
+            public Boolean apply(WebDriver webDriver) {
+                try {
+                    return until.get();
+                } catch (Exception ex) {
+                    return false;
+                }
+            }
+        });
+    }
```
martinmladenov, line 42: Is this the best way of preventing code duplication here?

Arraying: I can't think of a better solution off the top of my head.

mauricioaniche: Also looks good to me

PR #73, opened 2021-12-03
Add configuration options to skip Jacoco or Pitest

martinmladenov:
Added two methods in `RunConfiguration` - `skipJacoco()` and `skipPitest()`, which can be overriden in order to disable Pitest and Jacoco in exercises where it doesn't make sense to run them, such as exercises that rely on testing an external process.

### DIFF
--- a/src/main/java/nl/tudelft/cse1110/andy/config/RunConfiguration.java
+++ b/src/main/java/nl/tudelft/cse1110/andy/config/RunConfiguration.java
@@ -61,6 +61,10 @@ public abstract class RunConfiguration {
         return null;
     }
 
+    public boolean skipJacoco() {
+        return false;
+    }
+
     public boolean skipPitest() {
         return false;
     }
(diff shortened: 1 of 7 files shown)


src/main/java/nl/tudelft/cse1110/andy/config/RunConfiguration.java
```diff
@@ -61,4 +61,12 @@ public String successMessage() {
         return null;
     }
 
+    public boolean skipJacoco() {
```
mauricioaniche, line 64: do we need this?

why not, in the `getRunConfiguration().skipJacoco()`, we simply do ``return weight == 0``?

mauricioaniche (commented):
Suggestion here

src/main/java/nl/tudelft/cse1110/andy/config/RunConfiguration.java
martinmladenov: I would say it is nice to have the flexibility to disable it only when strictly necessary due to them breaking things (for instance, in Selenium tests, where there's no library class). I think those results should be shown to students even in cases where they are not taken into account for the final score. For instance, on an exercise whose final score is solely determined by meta tests and code checks, it may well still be useful for students to see code coverage information.

mauricioaniche (approved):
Good point. LGTM then!

PR #74, opened 2021-12-05
Create Selenium test that runs all exercises on WebLab

martinmladenov:
Added a Selenium test that runs all exercises and checks whether all spec tests get 100%.
The test performs recursive discovery of exercises on WebLab. It then runs all discovered exercises using the sample solution.
As the exercises are discovered using a JUnit generator and the test itself is a parameterised test, all exercises are treated as a separate instance of the test, so it's easy to see which exercise fails.

Furthermore, I changed the login mechanism, so logging in is only done once instead before every individual test. Also, I switched to using JavaScript to enter the solution code in order to improve test performance.

### DIFF
--- a/src/test/java/selenium/AllExerciseTests.java
+++ b/src/test/java/selenium/AllExerciseTests.java
@@ -3,29 +3,44 @@ package selenium;
 import org.junit.jupiter.params.provider.Arguments;
 import selenium.pageobjects.WebLabAssignmentListPage;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 public class AllExerciseTests extends WebLabSeleniumTestBase {
 
     protected static final String BASE_ASSIGNMENT_ID = "91800";
 
-    public Stream<Arguments> scrapeAssignmentUrls() {
-        return scrapeAssignmentUrls(WEBLAB_URL + String.format(WEBLAB_ASSIGNMENT_INFO_PATH, BASE_ASSIGNMENT_ID));
+    public Stream<Arguments> scrapeAssignmentIds() {
+        return scrapeAssignmentIds(WEBLAB_URL + String.format(WEBLAB_ASSIGNMENT_VIEW_PATH, BASE_ASSIGNMENT_ID));
     }
 
-    private Stream<Arguments> scrapeAssignmentUrls(String baseAssignmentUrl) {
+    private Stream<Arguments> scrapeAssignmentIds(String baseAssignmentUrl) {
         WebLabAssignmentListPage listPage = new WebLabAssignmentListPage(driver, baseAssignmentUrl);
         listPage.navigate();
 
         // Scrape programming assignments
-        Stream<Arguments> assignmentUrls = listPage.getAssignmentUrls().stream().map(Arguments::of);
+        Stream<Arguments> assignmentIds = listPage.getAssignmentUrls().stream()
+                .map(AllExerciseTests::getAssignmentId)
+                .map(Arguments::of);
 
         // Recursively scrape assignments in subdirectories
         for (String subdirectoryUrl : listPage.getSubdirectoryUrls()) {
-            Stream<Arguments> subdirectoryUrls = scrapeAssignmentUrls(subdirectoryUrl);
-            assignmentUrls = Stream.concat(assignmentUrls, subdirectoryUrls);
+            Stream<Arguments> subdirectoryAssignmentIds = scrapeAssignmentIds(subdirectoryUrl);
+            assignmentIds = Stream.concat(assignmentIds, subdirectoryAssignmentIds);
         }
 
-        return assignmentUrls;
+        return assignmentIds;
+    }
+
+    private static String getAssignmentId(String path) {
+        Pattern pattern = Pattern.compile("/assignment/(\\d+)/");
+        Matcher matcher = pattern.matcher(path);
+
+        if (!matcher.find()) {
+            throw new IllegalArgumentException("Could not parse path");
+        }
+
+        return matcher.group(1);
     }
 }
--- a/src/test/java/selenium/AllExerciseTests.java
+++ b/src/test/java/selenium/AllExerciseTests.java
@@ -1,0 +1,31 @@
+package selenium;
+
+import org.junit.jupiter.params.provider.Arguments;
+import selenium.pageobjects.WebLabAssignmentListPage;
+
+import java.util.stream.Stream;
+
+public class AllExerciseTests extends WebLabSeleniumTestBase {
+
+    protected static final String BASE_ASSIGNMENT_ID = "91800";
+
+    public Stream<Arguments> scrapeAssignmentUrls() {
+        return scrapeAssignmentUrls(WEBLAB_URL + String.format(WEBLAB_ASSIGNMENT_INFO_PATH, BASE_ASSIGNMENT_ID));
+    }
+
+    private Stream<Arguments> scrapeAssignmentUrls(String baseAssignmentUrl) {
+        WebLabAssignmentListPage listPage = new WebLabAssignmentListPage(driver, baseAssignmentUrl);
+        listPage.navigate();
+
+        // Scrape programming assignments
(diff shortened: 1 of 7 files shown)


src/test/java/selenium/AllExerciseTests.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class AllExerciseTests extends WebLabSeleniumTestBase {
+
+    protected static final String BASE_ASSIGNMENT_ID = "91800";
+
+    @BeforeAll
+    public void clearClipboard() {
+        // Clear clipboard in case there is something sensitive there
+        // so that its content does not get uploaded to WebLab if a test fails
+        Toolkit.getDefaultToolkit().getSystemClipboard().setContents(new StringSelection(""), null);
```
mauricioaniche, line 29: Why would this happen?

src/test/java/selenium/pageobjects/WebLabAnswerPage.java
```diff
@@ -0,0 +1,61 @@
...
+    public String getSolution() {
+        this.awaitPageLoaded();
+
+        // unlock answer if it is not yet unlocked
+        if (this.isDisplayed(unlockAnswerButton)) {
+            this.unlockAnswerButton.click();
+            this.awaitElementVisibility(this.solutionDiv);
+        }
+
+        this.solutionDiv.click();
```
mauricioaniche, line 36: Ahh, I see why you clear up the clipboard. 

I _would not_ do this this way. Just get the content via selenium, and put it in a variable. Why are you copying and pasteing?

mauricioaniche (changes requested):
I would avoid using copy and paste. Just get the content via selenium.

src/test/java/selenium/pageobjects/WebLabAnswerPage.java
martinmladenov: Because the solution isn't contained in a single element and the editor dynamically changes the contents of the elements, so there's no easy way to retrieve the solution apart from copying and pasting. It is indeed a dirty solution and it turns out it doesn't work on a headless browser, so we can't use it. I'll look into other ways to retrieve the solution, maybe via JavaScript.

src/test/java/selenium/pageobjects/WebLabAssignmentListPage.java
```diff
@@ -0,0 +1,51 @@
...
+                        .equals(type)
+                )
+                .map(tr -> tr
+                        .findElements(By.tagName("td"))
+                        .get(1)
+                        .findElement(By.tagName("a"))
+                        .getAttribute("href")
+                )
+                .collect(Collectors.toList());
+    }
```
martinmladenov, line 46: Is there a better way of doing this via the Page Object pattern?

src/test/java/selenium/pageobjects/WebLabAnswerPage.java
martinmladenov: @mauricioaniche Using JS worked flawlessly! :) In fact, using JS is much faster than typing the solution into the editor, so I implemented it in the submission code as well, which made the tests about 20% faster.

src/test/java/selenium/pageobjects/WebLabAnswerPage.java
```diff
@@ -0,0 +1,40 @@
...
+    public String getSolution() {
+        this.awaitPageLoaded();
+
+        // unlock answer if it is not yet unlocked
+        if (this.isDisplayed(unlockAnswerButton)) {
+            this.unlockAnswerButton.click();
+            this.awaitElementVisibility(this.solutionDiv);
+        }
+
+        String script = "return document.getElementById('solution').value;";
```
mauricioaniche, line 29: You can't get the element via selenium?

src/test/java/selenium/pageobjects/WebLabAssignmentListPage.java
mauricioaniche: What exactly? This looks good to me.

src/test/java/selenium/pageobjects/WebLabSubmissionPage.java
```diff
@@ -58,15 +59,22 @@ public void assessWithHints() {
...
 
-        this.solutionDiv.click();
-        Actions actions = new Actions(driver);
-        actions.keyDown(Keys.CONTROL);
-        actions.sendKeys("a");
-        actions.keyUp(Keys.CONTROL);
-        actions.sendKeys(Keys.BACK_SPACE);
-        actions.sendKeys(solution);
-        actions.perform();
+        String script = String.format("document.getElementById('%s').value = '%s';",
```
mauricioaniche, line 62: Same here, you can't do via selenium?

mauricioaniche (commented):
If not possible to do via selenium, id suggest a comment explaining why not

src/test/java/selenium/pageobjects/WebLabAnswerPage.java
martinmladenov: This element is invisible, so I don't think it's possible to get it directly via selenium.

mauricioaniche: https://stackoverflow.com/questions/20012442/using-selenium-can-i-get-the-text-of-an-invisible-element#24336401

Does this help?

src/test/java/selenium/pageobjects/WebLabAssignmentListPage.java
martinmladenov: Could we maybe have another class that describes a single `subAssignment` element? Something like the Page Object we have for the entire page, but only for one element with subelements, so that I don'thave to use `.findElements(By.tagName("td")).get(0).findElement(By.tagName("span")).getAttribute("data-original-title")` but instead I could do something like `subAssignment.getAssignmentType()`?

src/test/java/selenium/pageobjects/WebLabAnswerPage.java
martinmladenov: Thank you, this did the job. It seems like `getAttribute` works even if the element is not displayed.

src/test/java/selenium/pageobjects/WebLabSubmissionPage.java
martinmladenov: I tried set the `textContent` attribute, but couldn't find a way to do that directly via Selenium without the use of JavaScript. Is there a method similar to `getAttribute` but for setting an attribute? It doesn't seem to exist in the `WebElement` class.

src/test/java/selenium/pageobjects/WebLabAssignmentListPage.java
mauricioaniche: Sure, why not?!

Or do you think a simple private method does the job?

src/test/java/selenium/pageobjects/WebLabSubmissionPage.java
mauricioaniche: Hmm google also can't find one. So I guess js magic is required here. Please add a comment explaining that (so that our students, when they read the code, they understand why we did this way)

martinmladenov: Sure, I added some comments.

src/test/java/selenium/pageobjects/WebLabAssignmentListPage.java
martinmladenov: I added a class named WebLabAssignmentListElement to do this. However, it seems like it's not possible to use `@FindBy` in nested objects without additional libraries, so I created the page object a bit differently than the rest: [WebLabAssignmentListElement.java](https://github.com/cse1110/andy/blob/094e6a0abe9128da6856dbe494d90401b285cdc2/src/test/java/selenium/pageobjects/WebLabAssignmentListElement.java)

Is there a better way to achieve this?

mauricioaniche (approved):
LGTM!

I think this is super clean, and the lack of the findBy annotation in the nested class is not a problem.

mauricioaniche (approved):
LGTM!

I think this is super clean, and the lack of the findBy annotation in the nested class is not a problem.

mauricioaniche (approved):
LGTM!!

Code is super clean!

PR #86, opened 2021-12-21
Overload constructor

### LINKED ISSUE #83: Overload Andy constructor
Overload the Andy constructor in order to preserve backward compatibility after the changes introduced in #70 as those changes break the Maven plugin

martinmladenov:
Overload the Andy constructor in order to preserve backward compatibility after the changes introduced in #70 as those changes break the Maven plugin

Fixes #83


mauricioaniche (approved):
Breaking the maven plug-in is not a problem, as no one is using it. But the code looks good to me!

PR #87, opened 2021-12-21
Read submission metadata from arguments instead of environment variables

### LINKED ISSUE #84: Read submission metadata from arguments instead of environment variables
Related to PR #70

martinmladenov:
Related to PR #70

Fixes #84


mauricioaniche (approved):
I usually prefer to make all the parsing in the main method. And then only pass beautiful variables to private methods. This avoids passing the horrible and unstructured args array. But just a matter of personal taste. LGTM as is!

PR #89, opened 2021-12-27
Add external process meta tests

### LINKED ISSUE #76: Add external process meta tests
Add meta tests that can start other external processes and run the tests via junit to verify that at least one of them fails.

Make sure to kill the main external process before that so that the various external processes launched while assessing the code don't interfere with each other.

martinmladenov:
- [x] Make external process execution more flexible
- [x] Make `MetaTest` abstract
  - [x] Move current `MetaTest` functionality into `LibraryMetaTest`
  - [x] Extract evaluators into their own classes
  - [x] Verify refactoring the `MetaTest` class preserves backward compatibility with existing configurations
  - [x] Move MetaTest creation logic into a factory 
- [x] Implement `ExternalProcessMetaTest`
  - [x] Add unit tests for `ExternalProcessMetaTest`
- [x] Implement `RunExternalProcessMetaTestStep`
  - [x] Add integration tests with external process meta tests

Closes #76


src/main/java/nl/tudelft/cse1110/andy/config/MetaTest.java
```diff
@@ -1,140 +1,44 @@
...
-
-    public static MetaTest withStringReplacement(int weight, String name, String old, String replacement) {
-        MetaEvaluator evaluator = new StringReplacementEvaluator(old, replacement);
-        return new MetaTest(weight, name, evaluator);
-    }
+    String getNameAndWeight();
 
-    public static MetaTest withStringReplacement(String name, String old, String replacement) {
-        return withStringReplacement(1, name, old, replacement);
+    static MetaTest withStringReplacement(int weight, String name, String old, String replacement) {
```
mauricioaniche, line 13: Is the `default` visibility on purpose?

src/main/java/nl/tudelft/cse1110/andy/config/MetaTest.java
```diff
@@ -1,140 +1,44 @@
...
-        public String evaluate(String oldLibraryCode) {
-            List<String> lines = oldLibraryCode.lines().collect(Collectors.toList());
-            List<String> result = new ArrayList<>(lines.subList(0, this.start));
-            result.add(this.replacement);
-            result.addAll(lines.subList(this.end + 1, lines.size()));
-
-            String resultStr = String.join("\n", result);
-            return resultStr;
-        }
+    static MetaTest withExternalProcess(int weight, String name, ExternalProcess externalProcess) {
```
mauricioaniche, line 37: Ok, so the external process is a meta test? Let me read the rest.

src/main/java/nl/tudelft/cse1110/andy/execution/Context.java
```diff
@@ -84,6 +84,9 @@ public void setExternalProcess(ExternalProcess externalProcess) {
     }
 
     public void killExternalProcess() {
+        // Retrieve error messages before killing process
+        externalProcess.getErrorMessages();
```
mauricioaniche, line 88: retrieve but do nothing?

src/main/java/nl/tudelft/cse1110/andy/execution/metatest/implementations/LibraryMetaTest.java
```diff
@@ -0,0 +1,17 @@
+package nl.tudelft.cse1110.andy.execution.metatest.implementations;
+
+import nl.tudelft.cse1110.andy.execution.metatest.AbstractMetaTest;
+import nl.tudelft.cse1110.andy.execution.metatest.evaluators.MetaEvaluator;
+
+public class LibraryMetaTest extends AbstractMetaTest {
```
mauricioaniche, line 6: What is a "LibraryMetaTest"? I don't get it by its name.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunExternalProcessMetaTestsStep.java
```diff
@@ -0,0 +1,82 @@
...
+
+        int score = 0;
+        int totalWeight = 0;
+
+        try {
+            List<MetaTest> metaTests = runCfg.metaTests();
+            List<MetaTestResult> metaTestResults = new ArrayList<>();
+
+            for (MetaTest abstractMetaTest : metaTests) {
+                if (!(abstractMetaTest instanceof ExternalProcessMetaTest)) {
```
mauricioaniche, line 32: I was indeed thinking that the meta tests we have and this one are different things... and the need for an `instanceof` shows that. If you gotta ask the class about which concrete implementation it actually is, then, this means this is not a good abstraction anymore.

src/main/java/nl/tudelft/cse1110/andy/result/MetaTestsResult.java
```diff
@@ -1,33 +1,43 @@
...
     }
 
     public static MetaTestsResult empty() {
-        return new MetaTestsResult(false, 0, 0, Collections.emptyList());
+        return new MetaTestsResult();
+    }
+
+    public MetaTestsResult addResults(int score, int totalTests, List<MetaTestResult> metaTestResults) {
+        if (metaTestResults.size() > totalTests)
+            throw new RuntimeException("Number of meta tests does not match.");
```
mauricioaniche, line 30: Maybe add in the exception message something that makes us understand this is a pre-condition being broken?

mauricioaniche (commented):
This is a super long PR, so I may need to review it more times.

Besides my comments, my main concern is about us using `instanceof`. This is a smell that basically tells us that our abstraction is not good enough. It seems to me that the meta tests we had before and this new meta test are two different things, and shouldn't be on the same abstraction.

src/main/java/nl/tudelft/cse1110/andy/config/MetaTest.java
martinmladenov: That is in an interface - the visibility is actually public.

src/main/java/nl/tudelft/cse1110/andy/execution/Context.java
martinmladenov: When calling this method, the ExternalProcess class retrieves them from the process and stores them. On subsequent calls of the method, those already stored messages will be returned. That has to be done before killing the process since retrieving them from the process after that can be unreliable.

src/main/java/nl/tudelft/cse1110/andy/execution/metatest/implementations/LibraryMetaTest.java
martinmladenov: A LibraryMetaTest is a meta test that replaces code in the library. This is the meta test type that corresponds to the already existing meta tests (inserting lines, replacing a substring, etc.)

src/main/java/nl/tudelft/cse1110/andy/result/MetaTestsResult.java
martinmladenov: Sure, I will do it in a bit.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunExternalProcessMetaTestsStep.java
martinmladenov: The main reason I introduced this abstraction is that I wanted to make defining external process meta tests in the configuration just as simple as defining other meta tests:
```java
    @Override
    public List<MetaTest> metaTests() {
        return List.of(
                MetaTest.withExternalProcess(2, "home page does not load",
                        new CommandExternalProcess(
                                "java -jar petclinic-metatest1.jar",
                                "initialised")
                )
        );
    }
```

My initial idea was to have two separate strategies for evaliating meta tests - one which replaces the library and one which replaces the external process. This would eliminate the need for two separate steps. However, this is not possible without running the step twice - before and after terminating the original external process,  since library replacements have to be done before the original external process is killed. (Technically, Selenium assignments don't need to have a library, so it wouldn't make a difference even if it doesn't work for them, but maybe there could be Selenium assignments with library code in the future.)

I'll think about this in the next few days and see if I can figure out another way of executing meta tests with strategies.

src/main/java/nl/tudelft/cse1110/andy/execution/metatest/evaluators/StringReplacementEvaluator.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public String evaluate(String oldLibraryCode) {
+        String shiftedOld = this.old.replaceAll("(?m)^\\s+", "");
+        String shiftedOldLibraryCode = oldLibraryCode.replaceAll("(?m)^\\s+", "");
+        String replaced = shiftedOldLibraryCode.replace(shiftedOld, this.replacement);
+
+        if (replaced.equals(oldLibraryCode)) {
+            throw new RuntimeException("Meta test failed to find this text replacement:\n" + this.old);
```
martinmladenov, line 20: Just something I remembered: When I was converting assignments in September, there was a bug with this in previous versions of Andy and this runtime exception is actually suppressed later, so it simply causes the meta test to fail instead of raising a generic failure, which is a bit confusing. I'll see if there's a clean way of fixing that (if it hasn't been fixed already due to the refactoring).

src/main/java/nl/tudelft/cse1110/andy/config/MetaTest.java
mauricioaniche: Oh yeah, of course, I missed that.

src/main/java/nl/tudelft/cse1110/andy/execution/Context.java
mauricioaniche: So, I'd call it something else than `get...`. Maybe `extract...`?

src/main/java/nl/tudelft/cse1110/andy/execution/metatest/implementations/LibraryMetaTest.java
mauricioaniche: Ok, let's keep it like this. For a second, I thought of giving the name of what these meta tests do, in this case, they mutate the code under test. But let's see how it goes first before giving better names.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunExternalProcessMetaTestsStep.java
mauricioaniche: I think these two things will never be mixed. If it's selenium, then, it's a black box test and there won't ever be library code. Or vice-versa. Can you see one example of both existing together? And anyways, if this forces us to use `instanceof`, I'd say it's not worth merging them. `instanceof` is one of the biggest red flags we can have in code IMHO.

src/main/java/nl/tudelft/cse1110/andy/execution/metatest/evaluators/StringReplacementEvaluator.java
mauricioaniche: Please, open an issue about it. If there's an exception in the meta tests, we should fail the whole execution.

src/main/java/nl/tudelft/cse1110/andy/execution/step/RunExternalProcessMetaTestsStep.java
martinmladenov: I can't think of any use case where both will be used simultaneously.

I have an idea how we can remove `instanceof`. We could use the strategy pattern (in order to remove the need for two separate meta test executors) together with an `executionType` which will be different per meta test type - that execution type can be either `BEFORE_KILLING_EXTERNAL_PROCESS` or `AFTER_KILLING_EXTERNAL_PROCESS`. We can then call the meta test step twice in the pipeline - once before killing the external process, which would execute meta tests of the first type, and once after that, which would run the other meta tests. This would remove the need for `instanceof` and would instead move this logic to the meta test class itself.

mauricioaniche: The problem is that the "before killing the process" type makes no sense at all for the LibraryMetaTests. :/ The external process thingy is leaking to a place it shouldn't

martinmladenov: Should I then simply run all meta tests after the process is killed (which would work for external meta tests and wouldn't make any difference for library meta tests)? That way we don't need any such complicated logic, and the `instanceof` can easily be replaced by separating the execution strategies.

mauricioaniche: Not sure what you mean (running after the process is killed), but sure, give it a try! If it removes the need for instanceof I'm happier :)

martinmladenov: I fixed it by moving the execution logic to the meta test children themselves.

src/main/java/nl/tudelft/cse1110/andy/execution/metatest/evaluators/StringReplacementEvaluator.java
martinmladenov: Fixed in #91

PR #101, opened 2022-04-29
Fix score of 100/100 being achievable when some components are incomplete

### LINKED ISSUE #99: Score of 100/100 is possible even if not everything is completed
Grades >= 99.5 get rounded up to 100. 

This is an edge case: any grade between 99 and 100 should be rounded down to 99, otherwise students may get confused and think that they have completed an exercise even if they haven't actually passed everything.

For example:
```
Status: Done
Andy v0.27.1-85578d2 (2022-04-25T09:29:42+0000)

[...]

--- Assessment
Branch coverage: 20/21 (overall weight=0.10)
Mutation coverage: 27/27 (overall weight=0.30)
Code checks: 0/0 (overall weight=0.00)
Meta tests: 17/17 (overall weight=0.60)

Final grade: 100/100

    Super congrats!

[...]
```

Note the number of covered branches - only 20 out of 21 branches are covered, but due to the low weight of the branch coverage score, this branch is worth less than 0.5/100, so the final score becomes 99.5/100, which Andy incorrectly rounds up to 100/100.

martinmladenov:
# Description of the bug

Due to a bug, grades between 99.5 and 100 used to get rounded up to 100.

This is an edge case: any grade between 99 and 100 should be rounded down to 99, otherwise students may get confused and think that they have completed an exercise even if they haven't actually passed everything.

For example:
```
Status: Done
Andy v0.27.1-85578d2 (2022-04-25T09:29:42+0000)

[...]

--- Assessment
Branch coverage: 20/21 (overall weight=0.10)
Mutation coverage: 27/27 (overall weight=0.30)
Code checks: 0/0 (overall weight=0.00)
Meta tests: 17/17 (overall weight=0.60)

Final grade: 100/100

    Super congrats!

[...]
```

Note the number of covered branches - only 20 out of 21 branches are covered, but due to the low weight of the branch coverage score, this branch is worth less than 0.5/100, so the final score becomes 99.5/100, which Andy incorrectly rounds up to 100/100.

# How the problem was resolved
- Added a check in the grade calculation logic, where every grade between 99.5 and 100 would be rounded down to 99 instead of up to 100.
- Added tests to make sure this issue does not reoccur in the future.

Closes #99


src/main/java/nl/tudelft/cse1110/andy/grade/GradeCalculator.java
```diff
@@ -24,6 +24,11 @@ public int calculateFinalGrade(GradeValues gradeValues, GradeWeight weights) {
         if(finalGrade < 0 || finalGrade > 100)
             throw new RuntimeException("Invalid grade calculation");
 
+        // Grades between 99.5 and 100 should be rounded down to 99 instead of up
+        if (finalGrade == 100 && finalDecimalGrade < 1) {
```
mauricioaniche, line 28: Can we use the sum of integers (with all the points from the different criteria) rather than the float? That’d be a tad safer against yet another rounding error.

martinmladenov: I changed it to compare the scores of all components with a positive grade instead of the decimal grade

PR #128, opened 2022-08-29
Fix "no tests found" message being shown if generator method lookup fails

### LINKED ISSUE #117: No tests are found if generator method lookup fails
Instead of showing an exception, the message that no tests are found is shown if the generator method cannot be accessed (e.g. it is non-static or does not exist).

martinmladenov:
Instead of the "no tests found" message, the exception messages generated by JUnit or the tips from [JUnitUtils#simplifyTestErrorMessage](https://github.com/cse1110/andy/blob/71c986f0b05037696a8d605a1cb3131b079a7222/src/main/java/nl/tudelft/cse1110/andy/utils/JUnitUtils.java#L7) are shown when Andy finds no tests due to a reported test failure.

Fixes #117

### DIFF
--- a/src/main/java/nl/tudelft/cse1110/andy/writer/standard/StandardResultWriter.java
+++ b/src/main/java/nl/tudelft/cse1110/andy/writer/standard/StandardResultWriter.java
@@ -286,7 +286,7 @@ public class StandardResultWriter implements ResultWriter {
             return;
 
         l("\n--- JUnit execution");
-        if (tests.getTestsFound() == 0 && !tests.hasFailingMessage()) {
+        if (tests.noTestsFound()) {
             noTestsFound(tests);
         } else {
             l(String.format("%d/%d passed", tests.getTestsSucceeded(), tests.getTestsFound()));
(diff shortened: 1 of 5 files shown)


src/main/java/nl/tudelft/cse1110/andy/writer/standard/StandardResultWriter.java
```diff
@@ -286,7 +286,7 @@ private void printTestResults(UnitTestsResult tests) {
             return;
 
         l("\n--- JUnit execution");
-        if (tests.getTestsFound() == 0) {
+        if (tests.getTestsFound() == 0 && !tests.hasFailingMessage()) {
```
mauricioaniche, line 289: Two checks on the same `test` object, can we create a method inside this `UnitTestsResult` class, that encapsulates `tests.getTestsFound() == 0 && !tests.hasFailingMessage()`, and we give a nice name to it?

martinmladenov: Fixed

mauricioaniche (approved):
LGTM!

PR #132, opened 2022-09-18
Fix issue with test discovery and add a system test for an SQL exercise

### LINKED ISSUE #114: SQL tests can make test execution flaky
When running the Restaurants test with the code written below, Andy sometimes says that no tests were detected, and sometimes runs the test properly. It alternates between those behaviours seemingly at random, without any change in the code, and it happens both in the "Only tests" and "Assess" tabs. 
The bug has been reproduced by another student.
```    
@Test
    void getAverageRestaurantTest() {
        Restaurant r1 = new Restaurant(0, "a", "a", 6.4, "a");
        Restaurant r2 = new Restaurant(1, "b", "a", 6.5, "a");
        Restaurant r3 = new Restaurant(2, "c", "a", 7, "a");
        Restaurant r4 = new Restaurant(3, "d", "a", 8, "a");
        Restaurant r5 = new Restaurant(4, "e", "a", 8.1, "a");
        Restaurant r6 = new Restaurant(5, "f", "a", 3, "a");

        restaurantDao.save(r6);
        restaurantDao.save(r5);
        restaurantDao.save(r4);
        restaurantDao.save(r3);
        restaurantDao.save(r2);
        restaurantDao.save(r1);

        List<String> ret = List.of("c", "d");
        assertEquals(ret, restaurantDao.getAverageRestaurant());
    }
```

martinmladenov:
Added this test in an attempt to reproduce #114 

The issue with SQL tests is that there is a test template class in the library (which contains SQL-related test setup code). This class is the parent class of the student's actual test class.

This parent class also matched the rules for test class discovery - its name contained the substring `Test`. As Andy does not expect that, it picks an arbitrary class matching that condition, meaning that on each execution either the actual test class or this dummy class containing no test methods will randomly be picked to be run, leading to the observed flakiness.

I fixed that by renaming the template class and adding a check to Andy to throw an exception if multiple matching classes are found, instead of simply picking one to make sure this does not happen in the future.

Closes #114

### DIFF
--- a/src/test/java/integration/JUnitTestsTest.java
+++ b/src/test/java/integration/JUnitTestsTest.java
@@ -371,6 +371,18 @@ public class JUnitTestsTest {
             assertThat(result.getTests().hasTestsFailingOrFailures()).isFalse();
         }
 
+        // Multiple class names contain the substring "Test" -> test discovery fails
+        @Test
+        void multipleTestClassesDiscovered() {
+            Result result = run(Action.TESTS, "LibraryWithBadTestTemplate", "SolutionWithBadInheritance");
+
+            assertThat(result.hasFailed()).isTrue();
+            assertThat(result.hasGenericFailure()).isTrue();
+            assertThat(result.getGenericFailure().getExceptionMessage())
+                    .hasValueSatisfying(containsString(
+                            "java.lang.IllegalArgumentException: There are 2 classes containing the substring \"Test\""));
+        }
+
     }
 
     private static Condition<UnitTestsResult> failingTest(String nameOfTheTest) {
--- a/src/test/java/integration/JUnitTestsTest.java
+++ b/src/test/java/integration/JUnitTestsTest.java
@@ -226,6 +226,20 @@ public class JUnitTestsTest {
 
 
     @Nested
+    class SQLTests extends IntegrationTestBase {
+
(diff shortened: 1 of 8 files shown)


martinmladenov:
The test fails on the CI but passes locally, even in Docker... Odd.

martinmladenov:
image

src/test/java/integration/JUnitTestsTest.java
```diff
@@ -357,6 +371,18 @@ void forgotPropertyAnnotation() {
             assertThat(result.getTests().hasTestsFailingOrFailures()).isFalse();
         }
 
+        // Multiple class names contain the substring "Test" -> test discovery fails
+        @Test
+        void multipleTestClassesDiscovered() {
```
mauricioaniche, line 376: Maybe the name should say something about that "it should fail"

mauricioaniche (approved):
Minor comment. LGTM!

mauricioaniche:
Great catch, not an easy bug to spot!

src/test/java/integration/JUnitTestsTest.java
martinmladenov: Sure, I renamed it to `multipleTestClassesDiscoveredShouldThrowException`

PR #138, opened 2022-10-11
Update README

martinmladenov:
Added information about the Maven plugin.


mauricioaniche:
That's great. I see we don't have a manual on how to write the config files, right? Maybe we should have another issue for that.

martinmladenov:
@mauricioaniche Sure, I made an issue (#141).

In the meantime, should we merge this PR?

martinmladenov:
Added @Arraying to the contributor list in the readme.

PR #148, opened 2022-11-13
Display code snippet when there is a compilation error

### LINKED ISSUE #122: Show tip when there is a compilation error
> if error message is about lots of compilation errors and "did not find symbols", give a tip saying that for them to look at the first occurrence of the problem, and check whether it's valid java there"

martinmladenov:
Closes #122 

A code snippet is now shown for the first compilation error when the solution contains errors.

For example:

```
Andy v0.29-f72d378 (2022-11-13T13:28:24+0000)

We could not compile the code. See the compilation errors below:
- line 21: cannot find symbol
  symbol:   method isEqual(double[])
  location: class org.assertj.core.api.AbstractDoubleArrayAssert<capture#1 of ?>
    @MethodSource("generator")
    void unique(String description, double[] array, double[] expectedArray) {
-->     assertThat(MathArrays.unique(array)).isEqual(expectedArray);
    }


- line 25: cannot find symbol
  symbol:   variable tc2
  location: class delft.MathArraysTest
- line 33: cannot find symbol
  symbol:   variable tc2
  location: class delft.MathArraysTest

Andy took 0.9 seconds to assess your solution.
```


mauricioaniche:
This is great!

Can we also have unit tests for the code snippet generator? It feels it's also a good example to show students.

In the test code, maybe we can even have a comment with the specification-based reasoning! Again, to show up to students!

martinmladenov:
Sure, I'll add some tests. We could even turn it into an assignment :)

mauricioaniche (approved):
LGTM!

PR #149, opened 2022-12-15
Report detailed scores in results.xml

martinmladenov:
- [x] Use a proper XML library for creating the XML
- [x] Update to the new WebLab XML format which doesn't require repeating <testcase> nodes
- [x] Add tests for the new format
- [x] Report basic statistics (branch coverage, mutation coverage, number of passed code checks, number of passed meta tests) in the xml
- [x] Report detailed statistics for individual code checks and meta tests (score is 1 if passed, 0 if failed)
- [x] Add tests to verify statistics are reported correctly


mauricioaniche (approved):
LGTM! (Thanks for using a proper library to generate the XML!)

PR #150, opened 2022-12-24
Update highlights.json format

### LINKED ISSUE #146: Update highlights.json format
The highlights.json file should be renamed to `editor-feedback.json`.

Per the weblab documentation, it should be placed in the same directory as results.xml, and must be in the following format:

```json
[
    {
        "location": "SOLUTION", //one of ["LIBRARY", "SOLUTION", "TEST"]
        "startLine": 20, //int
        "endLine": 20, //int
        "severity": "Info", //one of ["Error", "Hint", "Info", "Warning"]
        "message": "100% coverage" //String
    },
    {
        "location": "LIBRARY",
        "startLine": 41,
        "endLine": 48,
        "severity": "Info",
        "message": "Some message"
    },
  [...]
]
```

martinmladenov:
Closes #146


andy/src/test/java/unit/writer/weblab/WebLabResultWriterTest.java
```diff
@@ -19,7 +19,7 @@
 import static org.assertj.core.api.Assertions.allOf;
 import static org.assertj.core.api.Assertions.assertThat;
 import static unit.writer.standard.StandardResultTestAssertions.*;
-import static unit.writer.weblab.WebLabHighlightsJsonTestAssertions.*;
+import static unit.writer.weblab.WebLabEditorFeedbackJsonTestAssertions.*;
 
 public class WebLabResultWriterTest extends StandardResultWriterTest {
```
mauricioaniche, line 24: Do we have a test for when there's no thing to show in the highlights?

mauricioaniche (approved):
LGTM!

Are we going to use this? Is the weblab team going to implement support for it?

martinmladenov:
@mauricioaniche 
> Are we going to use this? Is the weblab team going to implement support for it?

Yes, they said they're working on it.

andy/src/test/java/unit/writer/weblab/WebLabResultWriterTest.java
martinmladenov: This test class extends `StandardResultWriterTest`, so all tests for the standard result writer are also executed for the WebLabResultWriter. The tests there don't have any highlights set.

The tests there implicitly verify that the WebLabResultWriter doesn't crash when there are no highlights. But we don't have any assertions to check that nothing is written to the highlights file.

mauricioaniche: Ah, great! 

Is this assertion needed?

martinmladenov: I don't think so. I don't think there's any reasonable bug that could occur which would cause false data to be generated and written to the file if no such data is provided to the writer. And we already have tests that check that the writer doesn't throw an exception in this case.

mauricioaniche:
Great!

andy/src/test/java/unit/writer/weblab/WebLabResultWriterTest.java
martinmladenov: In fact, we also have system tests which would not generate any highlights. They check only the score and implicitly verify that Andy doesn't crash in this case.

PR #161, opened 2023-05-17
Tagging tests that are meant to run in CI

mauricioaniche:
In MR adds `@Tag` to tests that are more meant to be run in the CI. For example, the WebLab tests that require extra configuration. Tags make it easy for us to exclude them in local (via a simple JUnit configuration).

### DIFF
--- a/andy/src/test/java/selenium/SecurityExploitTests.java
+++ b/andy/src/test/java/selenium/SecurityExploitTests.java
@@ -1,5 +1,6 @@
 package selenium;
 
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 import selenium.pageobjects.WebLabSubmissionPage;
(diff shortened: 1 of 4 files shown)


andy/src/test/java/selenium/SecurityExploitTests.java
```diff
@@ -9,6 +10,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static unit.writer.standard.StandardResultTestAssertions.*;
 
+@Tag("ci")
 public class SecurityExploitTests extends WebLabSeleniumTestBase {
```
martinmladenov, line 14: Those tests aren't meant to run only in CI. Do they fail on your machine?

mauricioaniche: Still not working here. I’ll merge this anyways. This isn’t a test that I feel like we have to run locally, let the CI do it. Also, note that the tag only gets disabled if you manually do it, I didn’t change the maven script.

PR #182, opened 2023-05-30
Draft: Assignments

mauricioaniche:
This MR adds all the assignments to this repository. They are all maven sub-modules. This MR also contains a few other things:

* Brought back the libraries to include as it's required by the aws plugin
* Added compile everything to the CI
* Trying to bring the maven plugin back, as this would be the easiest way to run Andy, even from intellij


Arraying:
Relevant tweet (I wonder what he would have to say for 31k lines):
image

If you want my opinion: looks good! 😂

PR #196, opened 2023-06-02
Running Andy in all assignments

mauricioaniche:
This test runs Andy in all assignments and ensures that they all get a 100/100.


mauricioaniche:
I’ll check on Tuesday why there’s a fail when compiling + installing.

PR #205, opened 2023-06-07
Add a slow test so that we can track performance

mauricioaniche:
This test takes too long in Weblab, but locally it works in 30-35 seconds. Let's have this one as a way to track that none of our changes make it too slower.


mauricioaniche:
Closing this one down because it doesn't work that easily.

PR #211, opened 2023-06-11
Add more code checks

martinmladenov:


mauricioaniche (changes requested):
We should write tests for those :)

mauricioaniche:
@martinmladenov should we close this one?

martinmladenov:
I think it's useful to have these code checks too. I'll write some tests for them :)

mauricioaniche:
I just merged method in a class code check!

martinmladenov:
Ah, I see. Maybe in that case we can keep just the AnnotatedMethod check?

mauricioaniche:
Yes!

Issue #173, opened 2023-05-26
Check if the RunConfiguration is correct should be done before the steps


mauricioaniche:
If weights are wrong, we only check in the end, after compiling, running, and etc. We can move these checks to earlier in the process.

just-rares:
Hello!

mauricioaniche:
This issue is yours, @just-rares!

PR #212, opened 2023-06-13
Run configuration extra step

just-rares:
This pull request solves issue #173 

Added a method in the `GetRunConfigurationStep` that checks the weights of the RunConfiguration and asserts whether they are valid with all the information available at the moment.

Changes: 
* Check if the weights add up to 1.0
* Check if there is a non-zero weight at mutants but Pitest is disabled
* Check if there is a non-zero weight at line coverage but Jacoco is disabled

If the configuration is invalid, the default one is set and an exception is thrown.

I tests the functionality of this method in the `RunConfigurationTests` file, where I used both unit tests to check the requirements of the method under test, some mocking to allow for disabling Pitest and Jacoco and one integration test with a library file with invalid weights to assert the error is indeed thrown and caught in the `Result` file. 

Notes:
* I have left the default checks in the `GradeCalculator`. Some of the checks could not be done, such as the one with the grades. If it should be removed, please do let me know. 

Any suggestions are welcomed, I am not very confident about my implementation of this issue


Arraying:
Approved workflow run.

andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/GetRunConfigurationStep.java
```diff
@@ -25,16 +25,43 @@ public void execute(Context ctx, ResultBuilder result) {
             Class<?> runConfigurationClass = Class.forName(getConfigurationClass(ctx.getNewClassNames()), false, currentClassLoader);
             RunConfiguration runConfiguration = (RunConfiguration) runConfigurationClass.getDeclaredConstructor().newInstance();
 
+            if(!isValidRunConfiguration(runConfiguration)) {
+                // load default configuration
+                ctx.setRunConfiguration(new DefaultRunConfiguration(allClassesButTestingAndConfigOnes(ctx.getNewClassNames())));
+                throw new RuntimeException("The run configuration has invalid weights. Please check the file.");
```
mauricioaniche, line 31: Let's throw a new exception, maybe `AndyConfigurationException`. Andy should stop in case the configuration is invalid, so in this sense, there's no need to set a default configuration! 

Somewhere up we should catch this exception and generate an error output to the user. The message should clearly indicate this isn't the student's problem but the teacher's problem.

andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/GetRunConfigurationStep.java
```diff
@@ -25,16 +25,43 @@ public void execute(Context ctx, ResultBuilder result) {
...
     }
 
+    /**
+     * Checks whether the config is valid.
+     * - Sum of weights must be 1.0
+     * - If Jacoco / Pitest are skipped, the weight should be 0
+     * @param config The run configuration of the problem
+     * @return validity of the config weights represented as a boolean
+     */
+    public static boolean isValidRunConfiguration(RunConfiguration config) {
```
mauricioaniche, line 50: This method would look better inside the `RunConfiguration` class itself. Look at it: 99% of it is interacting with data from that class, this is a sign that this method should be there.

andy/src/test/java/integration/RunConfigurationTests.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class RunConfigurationTests extends IntegrationTestBase {
```
mauricioaniche, line 19: These are unit tests, and they will look nicer once the rule is inside the `RunConfiguration` and you can simply call a `config.isValid()` rather than making a static method call to somewhere else

andy/src/test/resources/grader/fixtures/Config/NumberUtilsAddConfigurationInvalid.java
```diff
@@ -0,0 +1,100 @@
+package delft;
+
+import nl.tudelft.cse1110.andy.config.RunConfiguration;
+import nl.tudelft.cse1110.andy.config.MetaTest;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Configuration extends RunConfiguration {
```
mauricioaniche, line 10: I think we still need an integration test, with a wrong configuration, and then, we assert that the produced `Result` contains the error message. (See other integration tests in the integration folder)

mauricioaniche (commented):
Hey, @just-rares! Thanks for your first MR!

I wrote somme comments! Would you mind adjusting the code based on them?

mauricioaniche:
I just noticed the `GradeWeights` class. It also does some validations there (doesn't allow the sum not to be 100%). Are you using it here?

andy/src/test/resources/grader/fixtures/Config/NumberUtilsAddConfigurationInvalid.java
just-rares: This should be the wrong configuration for one of the tests. Maybe I am understanding it wrong?

Here is the code for the test:
```java
    @Test
    void checkExceptionHandling() {
        Result result = run("NumberUtilsAddLibrary", "NumberUtilsAddAllTestsPass", "NumberUtilsAddConfigurationInvalid");
        assertThat(result.getGenericFailure().getExceptionMessage().toString().contains("The run configuration has invalid weights. Please check the file.")).isTrue();
    }
```
    
   Found in the test suite

mauricioaniche: Ah ok, I missed the `checkExceptionHandling`. So, I'd split the unit tests from this integration test. Don't put them on the same file, as this isn't how we have been doing :)

mauricioaniche:
I'm closing this one for now. We can always re-open once you are back to it!

PR #216, opened 2023-06-18
Alternate construct

adonev42:
With the normal syntax, if we encounter a match in method name, it will be right after when(...).
Example: when(mockedList.add("2")).thenReturn(true).
The order in which they will be processed is thenReturn(...) -> when(...) -> add(...)

With the other syntax we can encounter a match either before when or after "when" invocation
Example: `doNothing().when(mockedList).desiredMethod()`.
The order will be `desiredMethod()` -> `when(...)` -> `doNothing()`

Alternatively, we can also write doReturn(true).when(mockedList.add("1"))
the order will be `when(...)` -> `doReturn(...)` -> `add(...)`
So we check if we encounter match on the current command or on the previous

We only memorize the previous, if it is not a `when`, because if we encounter our matching `methodName`, before reaching when mode, we will miss it.
Example: `doNothing().when(mockedList).desiredMethod()`
As stated above if we memorize every previous element, after reaching when mode we would lose `desiredMethod()`

Furthermore, there is a list with allowed do* constructs. So if support for more methods is needed, they can simply be added to the list and tests must be written to ensure, that they are working.


andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockitoWhen.java
```diff
@@ -32,30 +33,59 @@ public MockitoWhen(String methodToSetUp, Comparison comparison, int expectedNumb
...
+                // With the other syntax we can encounter a match either before when or after it
+                // Example: doNothing().when(mockedList).someVoidMethod()
+                // the order will be someVoidMethod() -> when(...) -> doNothing()
+                // Alternatively, we can also write doReturn(true).when(mockedList.add("1"))
+                // the order will be when(...) -> doReturn(...) -> add(...)
+                // So we check if we encounter match on the current command or on the previous
                 numberOfCallsToWhen++;
+                inWhenMode = false;
+                previousMethodName = "";
+            } else if(previousMethodName.contains("do") && !methodName.equals("when")){
```
mauricioaniche, line 57: Do we need to be specific here, instead of `contains("do")`, more like "equals("doNothing")` or is that not a problem?

andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockitoWhen.java
```diff
@@ -32,30 +33,59 @@ public MockitoWhen(String methodToSetUp, Comparison comparison, int expectedNumb
...
+                // If we did not encounter a match, and we are at a do* construct
+                // it means that we are going to the next statement and if it starts with "when",
+                // we do not reset inWhenMode. We clear previous to avoid potential bugs
+                inWhenMode = false;
+                previousMethodName = "";
+            }  else if(!methodName.contains("do") && !methodName.equals("when")){
```
mauricioaniche, line 63: Do we this comparison with `when` and `do` in many places. Let's extract this to a private utility method like `isMockitoWhenMethod(...)`. Same for the `doNothing`.

andy/src/test/resources/codechecker/fixtures/MockitoWhenCalls.java
```diff
@@ -13,12 +15,14 @@ public class MockitoWhenCalls {
     void t1() {
         List<String> mockedList = mock(List.class);
 
+        Mockito.doThrow(IllegalAccessError.class).when(mockedList).toString();
```
mauricioaniche, line 18: Can you not change this fixture, but add a new one instead? This way we are pretty sure existing behavior is kept!

andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockitoWhen.java
mauricioaniche: I see it's because `doReturn` and `doNothing`. I think it's best to be super precise here. If you extract to a private method, it becomes easier to compare for both.

adonev42: I totally agree, I added a list, which contains some doWhen invocations. If there is any need to add more, it is very simple to do so, by adding them to the list.

andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockitoWhen.java
adonev42: I agree that these conditions are sloppy and redundant, and I managed to reduce them to one method.

andy/src/test/resources/codechecker/fixtures/MockitoWhenCalls.java
adonev42: I should have done this from the beginning. I restored their state and added additional tests

martinmladenov:
@adonev42 Thank you for your contribution!

It seems like this still fails our linting rules: `- visit/1[nl.tudelft.cse1110.andy.codechecker.checks.MethodInvocation] in class nl.tudelft.cse1110.andy.codechecker.checks.MockitoWhen has a total lines of code of 27`

mauricioaniche:
@martinmladenov We should put this method in the exception list. It crosses the limit because of the comments which are not ignored by codesheriff I guess. Maybe best is to put a real linter here instead of my toy project!

mauricioaniche:
@adonev42 can you add this class as an exception? See the `CodeSheriffTests`, and add the `MockitoWhen` class as an exception there! And we merge it right away, on time for this to be available to the exam! :)

adonev42:
Hello @mauricioaniche,
The method should be added as an exception

PR #217, opened 2023-06-20
Exam mode security guard

mauricioaniche:
This MR introduces a simple exam mode security guard.

If we are running in exam mode, we replace the current ctx's `RunConfiguration` to a default one that has no meta tests or code checks inside. 

We also delete the `Configuration.java` and the `Configuration*.class` compiled version. This means any attempt to read the file from the disk or to load it via reflection should fail.


Arraying:
Out of curiosity because I don't remember the pipeline too well -- is there a chance that the old `Configuration.class` containing the metatests gets classloaded before it is deleted? Is it possible that a representation of this class still exists in memory?

mauricioaniche:
It's definitely possible, @Arraying, that's what my follow-up commit handles. I reset the classloader now!

martinmladenov (changes requested):
The `RunConfiguration` can also have other configuration options, such as external process configuration, disabling pitest/jacoco, etc. If I understand correctly, the current code only copies the classes under test. We need to copy all other options too (except code checks and meta tests, of course).

martinmladenov:
Another comment: we remove the RunConfiguration, but would it perhaps be possible for someone to retrieve references to the previously instantiated MetaTest / CodeCheck objects?

mauricioaniche:
Good catch, @martinmladenov! I'll copy the other parameters from the RunConfiguration.

No, for multiple reasons. The attacker can have access to the `MetaTest` class/interface but never on the concrete one we have that's produced by the call to `metaTests()` or whatever in our concrete `Configuration` implementation. Given that their code can't get access to our Configuration class anymore, there's no chance.

mauricioaniche:
Done, @martinmladenov! Please, take a look at it again and merge it if ready. I think we need better Selenium integration tests. I'll open a ticket for that.

martinmladenov (approved):

PR #225, opened 2023-06-26
Fix Docker using an old branch version from cache

martinmladenov:
Docker was caching the `git clone` command, causing it to use an old branch version from cache. Fixed it by forcibly writing a file which changes every time a new commit is pushed (from the GitHub API), which should block the caching mechanism, as the image state would be different.

### DIFF
--- a/weblab-docker/Dockerfile
+++ b/weblab-docker/Dockerfile
@@ -19,10 +19,13 @@ SHELL ["/bin/bash", "-c"]
 # Copy the full repo to the docker image
 ARG ANDY_DIR=/home/andy/
 ARG ANDY_BRANCH=main
-RUN git clone -n https://github.com/SERG-Delft/andy.git && git -C ${ANDY_DIR} checkout ${ANDY_BRANCH}
+RUN git clone -n https://github.com/SERG-Delft/andy.git
+
+# Checkout the current branch
+WORKDIR ${ANDY_DIR}
+RUN git checkout ${ANDY_BRANCH}
 
 # Compile and install all dependencies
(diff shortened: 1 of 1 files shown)


weblab-docker/Dockerfile
```diff
@@ -16,13 +16,20 @@ RUN apt update \
 
 SHELL ["/bin/bash", "-c"]
 
-# Copy the full repo to the docker image
 ARG ANDY_DIR=/home/andy/
 ARG ANDY_BRANCH=main
-RUN git clone -n https://github.com/cse1110/andy.git && git -C ${ANDY_DIR} checkout ${ANDY_BRANCH}
```
mauricioaniche, line 22: Can’t we simply pass a dynamic variable here, like the latest_commit one you use? Git clone accepts a -c option, allowing you to pass a configuration. Would that simplify?

martinmladenov: What do you mean exactly? What would this configuration contain?

mauricioaniche: Forget it. I thought “latest commit” was a variable, but this is simply putting the content of the page to this file. Can’t we just use some docker flag to force it to ignore the cache and rebuild the image?

martinmladenov: There doesn't seem to be anything like that, unfortunately... We could clear the entire cache but this would affect the cache of other images unrelated to Andy too, so it's not ideal.

mauricioaniche: That sorts this discussion out then!

mauricioaniche (approved):
LGTM!

Issue #202, opened 2023-06-07
When running meta tests, we should skip Jacoco


mauricioaniche:
When running the meta tests, we don't need to instrument the code. The `RunConfiguration` already has a boolean `skipJacoco`, which makes the `InstrumentCodeForCoverageStep` to not do anything. We have to make sure this is set to true during meta tests (see `LibraryMetaTest`).

cashbreaker:
I would like to work on this one

mauricioaniche:
All yours, @cashbreaker!

mauricioaniche:
This issue doesn't exist. When running meta tests, we only add `Run JUnit` in the list of steps to be executed, so coverage isn't even there.

PR #230, opened 2023-07-05
When running meta tests, we should skip Jacoco

cashbreaker:
Solved issue #202 

I have checked the code and I think setting `skipJacoco` to true may be impossible (or too much hassle) since it is implemented as method that is overridden by user in config file.

This is the code that loads `RunConfiguration`
```
Class<?> runConfigurationClass = Class.forName(getConfigurationClass(ctx.getNewClassNames()), false, currentClassLoader);
RunConfiguration runConfiguration = (RunConfiguration) runConfigurationClass.getDeclaredConstructor().newInstance();
```

There is no easy way of overriding methods at runtime in Java. Therefore I think that second if in `InstrumentCodeForCoverageStep` is a simple and good solution. Which I have implemented in this branch
```
// Skip step if running meta test
        if (ctx.getAction().equals(Action.META_TEST)) {
            return;
        }
```

### DIFF
--- a/andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/InstrumentCodeForCoverageStep.java
+++ b/andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/InstrumentCodeForCoverageStep.java
@@ -4,6 +4,7 @@ import nl.tudelft.cse1110.andy.config.DirectoryConfiguration;
 import nl.tudelft.cse1110.andy.config.RunConfiguration;
 import nl.tudelft.cse1110.andy.execution.Context;
 import nl.tudelft.cse1110.andy.execution.ExecutionStep;
+import nl.tudelft.cse1110.andy.execution.mode.Action;
 import nl.tudelft.cse1110.andy.result.ResultBuilder;
 import nl.tudelft.cse1110.andy.utils.ClassUtils;
 import nl.tudelft.cse1110.andy.utils.FilesUtils;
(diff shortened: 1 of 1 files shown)


mauricioaniche (commented):
Thanks for the MR!

The simpler the better! Skipping e test if it's a meta test is good enough!

I think you have to skip it also in the Collect Coverage step that happens after JUnit runs. My guess is that tests are breaking right now (I just approved the pipeline to run, we will see! I'll check how to make this the default configuration).

Also, we should write some code here to ensure that the program does what we want. This is hard to test as we have no clear way to observe the execution of each meta test. In the RunMetaTest step, we only care about whether there's a failing test. The simple thing that comes to mind would be to add an assert there ensuring that coverage wasn't executed. This way, if something goes wrong in our implementation, the program will halt. Thoughts on this?

martinmladenov:
@mauricioaniche Does this step even run when executing meta tests? From what I see, only the RunJUnitTestsStep is executed there.

image


image

mauricioaniche:
Hmm, great catch, @martinmladenov. Does this mean we don't even need this PR?

cashbreaker:
I have just checked the code and as Martin showed at the moment Jacoco does not run when running meta test. And since user can't specify their own Steps therefore this PR solves nonexistent issue. I think we can just close it.

mauricioaniche:
Thanks for double checking, @cashbreaker. I'll close the PR and the issue!

PR #233, opened 2023-07-07
Ignore spaces at the ends of lines in the MetaTest StringReplacementEvaluator, fixes #156

### LINKED ISSUE #156: Ignore spaces at the ends of lines in the MetaTest StringReplacementEvaluator
`MetaTest.withStringReplacement` should ignore spaces not only at the beginning of lines but also at the end.

With the following library code:
```
      if (nums == null) {
          return false;
    }
```
and meta test:
```
                MetaTest.withStringReplacement("the nums is null",
                        """
                        if (nums == null) {
                          return false;
                        }
                        """,
                        "")
```
If there is a space after “{”, it will return an error:
```
java.lang.RuntimeException: Meta test failed to find this text replacement:
if (nums == null) {
  return false;
}
```
When the space is removed, it works properly.

Una865:
Added regex expression for capturing spaces on the end of the line.

Fixes #156


andy/src/main/java/nl/tudelft/cse1110/andy/execution/metatest/library/evaluators/StringReplacementEvaluator.java
```diff
@@ -12,8 +12,8 @@ public StringReplacementEvaluator(String old, String replacement) {
 
     @Override
     public String evaluate(String oldLibraryCode) {
-        String shiftedOld = this.old.replaceAll("(?m)^\\s+", "");
-        String shiftedOldLibraryCode = oldLibraryCode.replaceAll("(?m)^\\s+", "");
+        String shiftedOld = this.old.replaceAll("(?m)^\\s+|(?m) +$", "");
+        String shiftedOldLibraryCode = oldLibraryCode.replaceAll("(?m)^\\s+|(?m) +$", "");
```
martinmladenov, line 16: Perhaps it would be better to use `\s+` instead of a literal space since that would also cover tabs.

Una865: Thank you! Fixed it

mauricioaniche:
Just rebased and put the tests to run!

@martinmladenov this one is for you to merge once you feel it's ready!

mauricioaniche:
Hey, @Una865, it seems like the tests are breaking. I just re-ran the pipeline with the most recent version of the code. Would you mind taking a look?

martinmladenov:
Using `\s` in both sides of the regex caused blank lines to be removed by merging the previous and the next lines (e.g. `123\n\n456` would be turned into `123456`). This was causing an exception when running meta tests, and meta tests would be mistakenly marked as passing when they shouldn't be due to #256. (However, this still doesn't explain why having this error in meta tests doesn't trigger a generic failure in the meta test sub-result but simply causes individual test cases to fail...)

martinmladenov (approved):

mauricioaniche (approved):
If you trust this regex, I trust on it too, @martinmladenov !

martinmladenov:
I trust our tests and according to them, this regex works! :laughing:

mauricioaniche:
That's the answer I was looking for 😆

Issue #143, opened 2022-11-04
Andy tests fail on Windows


martinmladenov:
Running tests on Windows results in some failing tests. This occurs on GitHub actions, I have not tested it on an actual machine.

Could possibly be related to security manager issues.

This does not affect running tests via Maven: the Maven plugin works fine on Windows.

---

> It seems like some of our tests fail on Windows for some reason. But the weird part is that a few unit tests fail, but none of our system tests do. All of those unit tests just test basic features (not edge cases), so that should in theory also cause system tests to fail... :thinking:

_Originally posted by @martinmladenov in https://github.com/cse1110/andy/issues/142#issuecomment-1289016956_

PR #236, opened 2023-07-08
Improved performance of the github workflow 

### LINKED ISSUE #214: Make assignment tests faster
Currently we have 4 tests which run all of our assignments - on linux, windows, macos, and docker. They're really slow. We don't need to run all assignments on all 4 platforms. Instead, we should run only 3 on most of them.

We should have one test which does run all of them - preferably the Docker one.

We can also think about running all assignments in parallel in order to speed up the workflow.

alexcojocaru2002:
Closes #214 

In this merge request only one random test is run on docker from each category (seen in the python script) and the workflow has been split in one job for each category to improve performance. 

image


mauricioaniche:
We gotta review this one together with #238

Arraying:
@mauricioaniche I'm happy for you to just review and this one, and then I'll incorporate the changes in my branch.

alexcojocaru2002:
> @mauricioaniche I'm happy for you to just review and this one, and then I'll incorporate the changes in my branch.

Same for me. 😅

mauricioaniche:
@martinmladenov can you work on this one?

mauricioaniche (approved):
This looks good to me!!

@martinmladenov any thoughts? If not, feel free to merge it!

mauricioaniche:
@alexcojocaru2002 There's a small conflict now, as we've added checkstyle to the build. I think it should run before the tests, and if it breaks, we can fail fast the build. Can you make this change?

alexcojocaru2002:
> @alexcojocaru2002 There's a small conflict now, as we've added checkstyle to the build. I think it should run before the tests, and if it breaks, we can fail fast the build. Can you make this change?

Sorry for the late reply, I am in vacation for a bit as well 😅 I will update the changes based on the checkstyle today or tomorrow 👍 If there are any other things to add please let me know

martinmladenov:
What's the difference between `run_unit_tests` and `run_andy_tests`? Those both seem to do the same thing.

.github/workflows/tests.yml
```diff
@@ -28,20 +29,68 @@ jobs:
...
+        with:
+          path: ~/.m2
+          key: ${{ matrix.job.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ matrix.job.os }}-m2
 
       - name: Run Andy unit tests
         run: mvn test -pl andy -DexcludedGroups=assignments
 
-      - name: Run Andy in all assignments
-        run: mvn test -pl andy -Dgroups=assignments
```
martinmladenov, line 38: We shouldn't delete these tests

.github/workflows/tests.yml
```diff
@@ -28,20 +29,68 @@ jobs:
...
 
+      - name: Cache Maven packages
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2
+          key: ${{ matrix.job.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ matrix.job.os }}-m2
+          
       - name: Run Checkstyle
         run: mvn checkstyle:check
```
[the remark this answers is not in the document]
alexcojocaru2002: I agree, I will try to make a dependency for the build on it 👍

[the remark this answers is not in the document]
mauricioaniche: Just on one OS (Linux) is fine!

alexcojocaru2002:
Also @martinmladenov @mauricioaniche has there been any changes to the weblab tests ? They seem to fail now due to a missing dependency, I will see if I can fix it but maybe something changed in the last PRs that I should be aware of and I was not able to check😅

alexcojocaru2002:
> All tests should be passing! Maybe rebase with master ?

I fixed it locally by doing a clean install before the weblab tests, I am not sure why... I will try to see if they pass without it first

mauricioaniche:
Pipeline running!

alexcojocaru2002:
> Pipeline running!

Yes, seems like something changed in the last week that made the weblab tests not able to find the dependencies anymore. 😢

alexcojocaru2002:
> Pipeline running!

The pipeline gives some errors regarding the ContextDirector / Builder in the AndyOnWeblab class.

mauricioaniche:
We saw this error before in this PR. This doesn't happen in master. Can you double check your code is rebased?

mauricioaniche:
@martinmladenov im guessing it's the maven cache here.

@alexcojocaru2002 can you add a mvn install to the script? If you look at the history of the file, you will find it. I think we removed it quite recently, but my feeling is that it's needed!

alexcojocaru2002:
> We saw this error before in this PR. This doesn't happen in master. Can you double check your code is rebased?

They probably work because the jobs include the 'Compile and Install everything' step, which in the end seems necessary for weblab tests.

alexcojocaru2002:
> 

Yep, I will, and it should be done 👍

mauricioaniche:
For the weblab module, the Andy jar needs to be available. So, my guess would be that before running weblab tests, we need to do a mvn install -pl andy

alexcojocaru2002:
> For the weblab module, the Andy jar needs to be available. So, my guess would be that before running weblab tests, we need to do a mvn install -pl andy

I did the original clean install -Dskiptests for shorter time, it seems to work now 👍. Sorry for the large number of commits, I had to test different small changes

mauricioaniche:
Great to hear it works! @martinmladenov can you give it a final look and merge ?

.github/workflows/tests.yml
```diff
@@ -1,7 +1,31 @@
...
+
+      - name: Cache Maven packages
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2
+          key: ${{ matrix.job.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ matrix.job.os }}-m2
+
+      - name: Run checkstyle
+        run: mvn checkstyle:check
```
mauricioaniche, line 25: Can we add fail fast here? If checkstyle fails, let's not even run the tests.

.github/workflows/tests.yml
```diff
@@ -28,20 +52,70 @@ jobs:
           key: ${{ matrix.job.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ matrix.job.os }}-m2
 
-      - name: Compile and install everything
-        run: mvn clean install -DskipTests
-
       - name: Run Andy unit tests
         run: mvn test -pl andy -DexcludedGroups=assignments
 
+  run_andy_tests:
```
mauricioaniche, line 58: Can we call this assignment tests? When I read Andy tests, I think of Andy's unit tests.

.github/workflows/tests.yml
```diff
@@ -28,20 +52,70 @@ jobs:
...
+
+      - name: Cache Maven packages
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2
+          key: ${{ matrix.job.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ matrix.job.os }}-m2
+
+      - name: Compile and install everything
+        run: mvn clean install -DskipTests
```
mauricioaniche, line 118: Should we install only Andy? This will install all assignments as well, maybe it's not needed?

.github/workflows/tests.yml
alexcojocaru2002: I think they are skipped automatically by the use of .needs, but I will add it anyways to make sure

alexcojocaru2002: I think they are skipped automatically by the use of .needs, but I will add it anyways to make sure

.github/workflows/tests.yml
```diff
@@ -28,20 +54,70 @@ jobs:
...
+          path: ~/.m2
+          key: ${{ matrix.job.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ matrix.job.os }}-m2
+
       - name: Run Andy in all assignments
         run: mvn test -pl andy -Dgroups=assignments
 
-      - name: Run Weblab runner tests
-        run: mvn test -pl weblab-runner -DexcludedGroups=selenium
+  run_weblab_tests:
```
martinmladenov, line 91:
```suggestion
  run_weblab_runner_tests:
```
I'm changing it because the name is a bit confusing; we also have a different set of tests which interact with WebLab platform itself.

martinmladenov:
I just changed the and now the pipeline fails. Seems like it can no longer resolve dependencies for some reason?

.github/scripts/assignments_test_docker.py
```diff
@@ -21,50 +21,51 @@ def get_directories(basedir):
...
-        # Update the `pipeline_failed` variable.
-        if score != 100:
-            print(output)
-            pipeline_failed = True
-
-        if expected_andy_version not in andy_version:
-            print(f'Error: Unexpected Andy version {andy_version}, expected {expected_andy_version}')
-            pipeline_failed = True
+    #for assignment_dir in get_directories(category_dir):
+    assignment_dir = random.choice(get_directories(category_dir));
```
martinmladenov, line 25: Would it perhaps be nice to run more than one assignment here? The Docker tests currently take about 2.5 minutes and this is much shorter than the slowest task, which takes 5.5 minutes. I think it might be nice to run 3 or 4 assignments per category - this won't affect the duration of the overall pipeline as different tasks runs in parallel. @mauricioaniche what do you think?

martinmladenov:
Is there a way to reduce code duplication in the workflow configuration file? I see that every task has the steps "Checkout the repository",  "Set up JDK 17" and "Cache Maven packages". Is there a way to define them only once in the file and make them apply to all tasks?

.github/workflows/tests.yml
```diff
@@ -28,20 +54,70 @@ jobs:
...
 
+  run_assignment_tests:
+    needs: checkstyle
+    runs-on: ${{ matrix.job.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        job:
+          - os: ubuntu-22.04
+          - os: macos-12
```
martinmladenov, line 68: Should we also run tests on Windows? #143 is fixed now.

martinmladenov:
One of the steps in the Docker task is called "Run the reference solutions of all assignments and verify the scores are 100/100". This is no longer correct, it should say "some assignments" or something similar.

.github/scripts/assignments_test_docker.py
mauricioaniche: Let's start with one and we can always open up another PR, this one is going too long!

.github/workflows/tests.yml
mauricioaniche: Let's do that in another PR too!

mauricioaniche:
> I just changed the and now the pipeline fails. Seems like it can no longer resolve dependencies for some reason?

Everything seems green here! Shall we merge this one?

martinmladenov:
It seems to pass now. Not sure what happened there. We can merge, but I left some more comments: see above.

martinmladenov:
I'll turn those comments into discussions so that they show up more clearly, this PR is very long now.

mauricioaniche:
We can reduce duplication in another MR to! I'd say we merge this one as soon as pipeline passes!

martinmladenov:
We also need to change the configuration before we can merge this one. I'll do it now.

martinmladenov:
@mauricioaniche That is so weird. Jobs seem to fail in the fork but pass on our repo https://github.com/alexcojocaru2002/andy/actions/runs/5615314967/job/15215354923

martinmladenov (approved):

alexcojocaru2002:
> Is there a way to reduce code duplication in the workflow configuration file? I see that every task has the steps "Checkout the repository", "Set up JDK 17" and "Cache Maven packages". Is there a way to define them only once in the file and make them apply to all tasks?

Unfortunately github workflow is a bit behind compared to its competitors when it comes to code duplication. There is not option to use yaml anchors which would ve made this way shorter in terms of code duplication. I thought of the file option but there is no way to take the environment of a job and continue it that i know of.

alexcojocaru2002:
> We can reduce duplication in another MR to! I'd say we merge this one as soon as pipeline passes!

Can be done in the one that should fix windows for pipeline, maybe, since it s going to add even more duplicated code

martinmladenov:
I see, thanks for the explanation and for your help with improving our CI! :)

PR #237, opened 2023-07-08
Reduce verbose test sysouts

cashbreaker:
Solves #207
I decided that simple naming scheme of index and name of parametrized test is good solution. It requires single line of code and if test fails it is easy to run it locally in IntelliJ and identify the problem.

Here are new outputs. I have changed the ones with long lines or containing line breaks.
Screenshot 2023-07-08 at 20 21 24
Screenshot 2023-07-08 at 20 21 39
Screenshot 2023-07-08 at 20 23 14


mauricioaniche (approved):
LGTM! Thanks for improving our CI!

PR #238, opened 2023-07-08
Run assignments in Docker for pull requests

### LINKED ISSUE #220: Make Docker tests work in forks/pull requests
Currently Docker assignment tests only run on pushes, not on pull requests.

Furthermore, the Dockerfile has the repo URL hardcoded. This causes it to break when it runs in a fork, as commits in forks are not present in the upstream repo. The URL should default to this one but it should be set in the CI via environment variables. GitHub provides such environment variables which we can use.

Arraying:
This encompasses:
- Assignments will now run in Docker when pull requests are created. If I create a pull request from `foo/serg-andy` to `SERG-Delft/andy`, then the assignments will be run, in Docker, on the `HEAD` of the head repository (here: `foo/serg-andy`). Pushes remain unaffected, the assignments will run, in Docker, on the `HEAD` of what is pushed.
- Refactored the pipeline a bit to facilitate this.
- Fixed a bug where assignments would not run if your fork's name is not `andy`.
- Removed `assignments_test.py` as it is no longer used (and suffers from the above bug), in consultation with @martinmladenov.

The comments self-explanatory, if you have any questions, let me know. Surprisingly, there is no way to set the [`[directory]` option when cloning](https://git-scm.com/docs/git-clone) in the checkout action (the Python script needs the repository location). This is poor design on behalf of GitHub actions, and I work around this by injecting the repository name excluding the owner alongside the commit hash.

Closes #220.


.github/workflows/assignments_test_docker.yml
```diff
@@ -21,12 +26,14 @@ jobs:
...
-          BUILD_ARGS: "--build-arg ANDY_BRANCH=${{ github.sha }}"
+          BUILD_ARGS: "--build-arg ANDY_BRANCH=${{ env.GHA_SHA }} --build-arg ANDY_REPO=${{ env.GHA_REPO }}"
         run: |
           make
 
       - name: Run the reference solutions of all assignments and verify the scores are 100/100
         env:
-          COMMIT_HASH: ${{ github.sha }}
+          # github.repository contains owner, we don't want that
+          REPO_NAME: ${{ github.event.repository.name }}
```
martinmladenov, line 36: Instead of using the repo name, could we perhaps set the `path` option of the checkout action to "andy" so that it is always "andy" even when the fork name is different?

martinmladenov, line 36: Instead of passing the repository name to the script in order to find the correct folder, could we use the "path" option of the checkout action to force it to always clone the repo to a directory called "andy" ? https://github.com/actions/checkout#usage

martinmladenov:
(Ignore the failing tests, it's not your fault, the docker image broken on the main branch :see_no_evil: )

.github/workflows/assignments_test_docker.yml
Arraying: I understood `path` to be the working directory in which `git clone` is issued because the documentation is abhorrent. It is possible to change it, but I feel this way is more robust -- we should not be relying on hardcoding `andy` anywhere really, in my opinion. Let me know what to do.

martinmladenov: I see, let's keep it as is then. Thanks!

martinmladenov (approved):

PR #239, opened 2023-07-09
Resolve "Create a proper builder for Context"

### LINKED ISSUE #191: Create a proper builder for Context
The `Context` class is built in many different places of the code. In some cases, we use the factory methods, in other cases we use the factory methods + extra setters. We gotta give it a proper builder, and then, get rid of the setters we don't need.

St1p42:
Closes #191 

### Builder
The Context class now has a builder `ContextBuilder` with setters and build() method. Setters not used during the construction of Context in the existing code are not present there, so if there are more cases where Context is created - these setters will have to be added. 
builder

### Director
Instead of calling builder.setA().setB().setC().build() to get a context I decided to have still some premade methods that are now in the class `ContextDirector`. Inside these construction methods, I also return the Context object to avoid using builder from the client code directly. I replaced all of the occurrences of previous initialization with these construction methods, though I am not sure if a director is necessary (I guess builder.setA().setB().setC().build() also works fine). Currently, it has `constructBase` (accepting action + directory config as parameters),
`constructWithLibraries` (accepting action + directory config + libraries list), and `constructWithClassLoader` (accepting action + directory config + libraries list + class loader).

### Other changes
Unnecessary setters in the `Context` class are now removed, though the ones which are used outside of initialization (for example when the setter is applied after passing initialized context to another method) are still there. I also created some tests for Directors' construction methods just in case.  Besides a folder Context was created to put there Context class + builder + director. So if there are too many file changes in this pull request, it is because of the fact that all imports of Context in the project were changed (since now it has a different path).


mauricioaniche:
I have to review this one on the computer!! I'm on vacation right now, so I'll do it next week!!

Thanks for the PR!

mauricioaniche (approved):
This looks like a great refactoring to me!

Tagging @martinmladenov so that he's also aware of it. Martin, if you have no comments, feel free to merge it!

Issue #224, opened 2023-06-26
Fix repository references


martinmladenov:
In various places in the repo we use the old repository URL (`cse1110/andy`) instead of the new one (`SERG-Delft/andy`). We should fix that, as the old one will stop working at some point.

Zakrok09:
Hello!

@sandemiroren1 and I would like to work on this issue.

martinmladenov:
Sure, I assigned you to the issue. GitHub doesn't let me assign @sandemiroren1 though, since he didn't comment on the issue.

sandemiroren1:
@martinmladenov it should be fixed now :D

PR #248, opened 2023-07-19
CHORE: References

sandemiroren1:
In this Pull Request we fix various references to the old URL of the repository and update them to the newer ones. 

Closes Issue #224
Worked on by @sandemiroren1 and @Zakrok09.


sandemiroren1:
We have changed all cse1110/andy references. There are also some places using the reference cse1110/* while linking to a repository should we also change those?

martinmladenov:
@sandemiroren1 We previously had multiple repositories (for Maven, the docker image, and the assignments), but we merged all of those into this repo. Those should now point to the relevant directory in this repo.

Zakrok09:
This should be all references updated, including to this repo and the `assignments` folder. 
The only file that is not updated is the [assignments_test.yml](https://github.com/SERG-Delft/andy/blob/main/.github/old/assignments_test.yml) located in a folder `old` inside `.github`. I suppose this file wouldn't need this update.

martinmladenov (approved):
LGTM!

PR #250, opened 2023-07-20
Add config setting for max discard ration to 2

### LINKED ISSUE #246: Extra configuration for jqwik
We should add the extra configuration `jqwik.maxdiscardratio.default = 2`. 

This will reduce the maximum discard ratio of JQWik. This will help us in quickly rejecting solutions that have bad data generation procedures.

(See #245 as this should also be implemented in pitest)

Zakrok09:
This PR adds another configuration step inside the jqwik config inside [RunJUnitTestsStep](https://github.com/SERG-Delft/andy/blob/main/andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/RunJUnitTestsStep.java) that sets the maximum discard ratio to 2.

Additionally, this PR adds a new Solution fixture to check for discarding of a "correct" solution that uses Assume to filter out disproportionately big number of entries and changes the `testMultiplePropertyTestsFailing` method and relating fixture to test the scenario if a correct property with bad assumption filtering fails in the context of other failing property-based tests.

Closes #246

### DIFF
--- a/andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/RunJUnitTestsStep.java
+++ b/andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/RunJUnitTestsStep.java
@@ -45,6 +45,7 @@ public class RunJUnitTestsStep implements ExecutionStep {
                     .configurationParameter("jqwik.reporting.usejunitplatform", "true")
                     .configurationParameter("jqwik.tries.default", "100")
                     .configurationParameter("jqwik.shrinking.default", "OFF")
+                    .configurationParameter("jqwik.maxdiscardratio.default", "2")
                     .configurationParameter("jqwik.database", FilesUtils.createTemporaryDirectory("jqwik").resolve("jqwik-db").toString())
                     .build();
             launcher.execute(request);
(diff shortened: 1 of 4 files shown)


mauricioaniche (changes requested):
Thanks for the PR!

It'd be great to have a test that shows that the test fails if discard ratio is high!

Zakrok09:
Hello again :wave: 
I added a new Solution fixture to check for discarding of a "correct" solution that uses `Assume` to filter out disproportionately big number of entries:

```java
    @Property
    void testNoElementInWholeArrayWithAssumption(
            @ForAll @Size(value=10) List<@IntRange(min = -2100000000, max = 2100000000) Integer> numbers,
            @ForAll int valueToFind) {

        Assume.that(valueToFind > 2100000000 || valueToFind < -2100000000);

        int[] arr = convertListToArray(numbers);
        assertEquals(-1, ArrayUtils.indexOf(arr, valueToFind, 0));
    }
```

I chose the numbers `2100000000` to minimize the risk of flakiness in the test. It could be too unrealistic however.

Additionally, I added this same method to a fixture that runs multiple failing properties and asserted that it fails too, in order to test this scenario too.

martinmladenov (approved):

mauricioaniche (approved):
LGTM!

PR #258, opened 2023-07-29
Replace external process by simple unzip step 165

NeilAlwani:
This pr will add unzipStep so that zipped files are unzipped in the working dir.


andy/src/main/java/nl/tudelft/cse1110/andy/config/SecureExamRunConfiguration.java
```diff
@@ -30,13 +26,14 @@ public Mode mode() {
     }
 
     @Override
+    @SuppressWarnings("DoubleBraceInitialization")
     public Map<String, Float> weights() {
-        return new HashMap<>(ImmutableMap.of(
```
mauricioaniche, line 34: I would expect this to break with errorprone! `ImmutableMap.of(...)` is how it should be now, I think!

PR #259, opened 2023-07-31
setting up Mock annotation to be used

### LINKED ISSUE #166: Make @Mock annotation available
Students should be able to use `@Mock` annotation from Mockito.

AtillaColak:
I updated the tests accordingly and they passed. They were some irrelevant failures on my local about weblab credentials not logged in, so I disregarded it since it's not related to this error. Please let me know if more changes have to be done to keep consistency. This MR closes #166

### DIFF
--- a/andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockClass.java
+++ b/andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockClass.java
@@ -1,6 +1,9 @@
 package nl.tudelft.cse1110.andy.codechecker.checks;
 
 import org.eclipse.jdt.core.dom.MethodInvocation;
+import org.eclipse.jdt.core.dom.Annotation;
+import org.eclipse.jdt.core.dom.FieldDeclaration;
+
 
 /**
  * Checks whether the class was mocked via Mockito.mock(..).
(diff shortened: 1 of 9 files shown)


mauricioaniche (approved):
This looks great to me! Super neat and clean code. Congrats!

I wouldn't expect tests to fail in local. Are they marked as CI tests? They shouldn't run locally. I'll put the pipeline to run here, let's see if everything passes!

andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockClass.java
```diff
@@ -38,6 +41,21 @@ public boolean visit(MethodInvocation mi) {
...
+    @Override
+    public boolean visit(FieldDeclaration fd) {
+        boolean hasMockAnnotation = fd.modifiers().stream()
+                .anyMatch(m -> m instanceof Annotation &&
+                        ((Annotation)m).getTypeName().getFullyQualifiedName().equals("Mock"));
+
+        // If the field is annotated with @Mock, check if it's the class we are interested in
+        if (hasMockAnnotation) {
+            String className = fd.getType().toString();
+            classWasMocked = className.contains(classToBeMocked);
```
martinmladenov, line 53: If I understand the code correctly, `classWasMocked` could be erroneously set to false if there are multiple usages of `@Mock`.  For example, the following could happen:
1. A search for a mocked instance of `Set` is started
2. `@Mock Set<Integer> a;` is visited,  `classWasMocked` is set to true
3. `@Mock List<String> b;` is visited,  `classWasMocked` is set to false
4. The final returned value is false

I see that the other visitor prevents this problem by only executing the logic if `classWasMocked` is still false.

martinmladenov (changes requested):

andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockClass.java
AtillaColak: Yeah, you're right. To handle that, I should add a `(!classWasMocked)` conditional. I will update the code and tests respectively.

AtillaColak: Updates committed and new test passes.

martinmladenov (approved):

andy/src/main/java/nl/tudelft/cse1110/andy/codechecker/checks/MockClass.java
martinmladenov: Thanks! :)

PR #260, opened 2023-08-01
exception handling Meta Tests

### LINKED ISSUE #227: If there's a compilation error in a meta test, inform the teacher
Right now, if there's a compilation error due to a badly written meta test, Andy just considers that meta test to fail. That's not correct. If the meta test makes the code to not compile, we should alert the teacher right away.

If during _EXAM_ mode (hopefully never), just show a message. In any other mode, paste the entire meta test + compilation error, so it gets easier to identify the problem.

Changes most likely should happen here: https://github.com/SERG-Delft/andy/blob/main/andy/src/main/java/nl/tudelft/cse1110/andy/execution/metatest/library/LibraryMetaTest.java

AtillaColak:
I initially tried to work with compilation exceptions in the `LibraryMetaTest` but as there was already exception handling in `RunMetaTestsStep` I made the changes there.  I was not sure about how to alert the teacher if there was a compilation exception so I worked with console logging. Please Let me know if something else was intended with "alerting the teacher".

Another thing to note is that when I made the changes in `LibraryMetaTest` as errors would already be handled here, in the `RunMetaTestsStep` we would get failing tests that were initially meant to be there for error handling. So this was another reason for why I worked with `RunMetaTestsStep` folder instead. 

I did not delete the other changes, just in case, only commented them out. So when we reach a final decision on where to handle errors, I will delete those commented lines.  

Closes #227

### DIFF
--- a/andy/src/main/java/nl/tudelft/cse1110/andy/config/SecureExamRunConfiguration.java
+++ b/andy/src/main/java/nl/tudelft/cse1110/andy/config/SecureExamRunConfiguration.java
@@ -39,6 +39,10 @@ public class SecureExamRunConfiguration extends RunConfiguration {
         ));
     }
 
+    public List<MetaTest> metaTests() {
+        return Collections.emptyList();
+    }
+
     @Override
     public List<String> classesUnderTest() {
         return Collections.unmodifiableList(classesUnderTest);
--- a/andy/src/main/java/nl/tudelft/cse1110/andy/execution/metatest/library/LibraryMetaTest.java
+++ b/andy/src/main/java/nl/tudelft/cse1110/andy/execution/metatest/library/LibraryMetaTest.java
@@ -30,44 +30,97 @@ public class LibraryMetaTest extends AbstractMetaTest {
     }
 
     @Override
-    public boolean execute(Context ctx, DirectoryConfiguration dirCfg, RunConfiguration runCfg) throws Exception {
-        /* Get the student solution, which we will run for each meta test */
-        String solutionFile = findSolution(dirCfg.getWorkingDir());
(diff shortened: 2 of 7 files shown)


mauricioaniche:
Hi, @AtillaColak, this is a hard MR to review, as I can't compare the previous code with the new one. I think it's easier if you don't have all the comments there, so that git diff can do its job.

What exactly are you trying to achieve? Which issue is this about?

AtillaColak:
> Hi, @AtillaColak, this is a hard MR to review, as I can't compare the previous code with the new one. I think it's easier if you don't have all the comments there, so that git diff can do its job.
> 
> What exactly are you trying to achieve? Which issue is this about?

This is an MR for the issue number #227. I've removed the unnecessary comments so it should be easier to review now.

martinmladenov (changes requested):
`System.out` and `System.err` are not reported by Andy; a good way to inform the teacher would be to use GenericFailures with an appropriate exception message (`result.genericFailure(this, ...)`).

Could you also write a test to verify that your changes work as expected?

AtillaColak:
> 

sure I'll get to it tonight

AtillaColak:
@martinmladenov Hey Martin could you take a look at this? For some reason, the test I wrote for the exam scenario fails and returns false for the "there is genericfailure" check. I would not want to disturb you if I could have solved it myself but I tried for some hours and was still at a dead end. Maybe you have an idea of what I could be obviously missing, so that I can fix it.

martinmladenov:
@AtillaColak 
Are you sure that your changes work correctly? If I'm not mistaken, in the current production version, if there's a compilation error due to a badly written meta test, Andy considers that meta test to fail (i.e. there's no generic failure raised). Your changes add an extra condition for exam mode in the part that raises a generic failure, but I don't think this part of the code gets triggered when a meta test encounters a compilation error.

PR #261, opened 2023-08-07
Add more assignment tests in docker 

### LINKED ISSUE #251: Run more than one assignment in Docker tests
It would be nice to run more than one assignment in Docker CI tests. The Docker tests currently take about 2.5 minutes and this is much shorter than the slowest task, which takes 5.5 minutes. We should run 3 or 4 assignments per category - this won't affect the duration of the overall pipeline as different tasks runs in parallel.

https://github.com/SERG-Delft/andy/pull/236#discussion_r1269839638_

alexcojocaru2002:
The aim of this MR is to add more random assignment tests (4) in docker, as per discussed in #236 

Closes #251

### DIFF
--- a/.github/scripts/assignments_test_docker.py
+++ b/.github/scripts/assignments_test_docker.py
@@ -22,6 +22,7 @@ os.makedirs(test_dir)
 pipeline_failed = False
 for category_dir in get_directories(home_dir):
     for assignment_dir in random.choices(get_directories(category_dir), k=4):
+        
         # Remove the contents of the output directory.
         os.system(f'rm -r {output_dir}')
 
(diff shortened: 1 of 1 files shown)


alexcojocaru2002:
Also maybe good to note the docker workflow is a bit flaky, not sure why
Between these 2 runs there is no change
image

Arraying:
The GH API sometimes rejects requests with invalid user agents, my guess is the `ADD` fell back to the default HTTP headers which GitHub was unhappy with. @martinmladenov this way of cache invalidation is very strange, perhaps we should create an issue to improve it. Specifically, `ADD` supports Git repositories natively, mayhaps we can replace the cache-buster and clone with a singular repository `ADD`.

martinmladenov:
@Arraying Interesting, I didn't know that, thanks. I saw a stackoverflow post which stated that downloading commit info to break the cache is the best solution. Let's make a new issue.

martinmladenov (approved):

PR #263, opened 2023-08-17
Run CI Tests on Windows

### LINKED ISSUE #218: Make pipeline to work on Windows
It seems we have a failure when running in Windows. Apparently when deleting the temporary directory we create during the tests.

See logs: https://github.com/cse1110/andy/actions/runs/5327003222/jobs/9649801907?pr=217

I'm disabling Windows for now.

### LINKED ISSUE #253: Run CI tests on Windows
We currently run CI tests on Linux and macOS. We should also run them on Windows.

https://github.com/SERG-Delft/andy/pull/236#discussion_r1269843286

alexcojocaru2002:
Closes #218.

Closes #253.

The aim of this PR is to add windows as an OS option when running the CI and to fix it in case the previous errors are still existent.


martinmladenov:
@alexcojocaru2002 Regarding #218: Did it turn out there was no bug after all? Does the temp directory get properly deleted?

alexcojocaru2002:
> @alexcojocaru2002 Regarding #218: Did it turn out there was no bug after all? Does the temp directory get properly deleted?
I think the bug fixed itself with the new additions to the workflow files... I can't see the logs anymore so I am not sure which temporary file it was, but I would assume yes since the actions are running fine.

martinmladenov (approved):

PR #264, opened 2023-08-17
Fix CI

martinmladenov:
This PR fixes failing CI jobs due to unavailable dependencies. 

Added the `-am` option which makes Maven work smarter by building dependencies automatically. (We were already using this in the Docker image but not in the test step in the CI.) Removed an unnecessary installation step as andy is now built automatically.


mauricioaniche (approved):
LGTM!

PR #267, opened 2023-09-02
Exammode configuration skipJacoco / skipPitest

### LINKED ISSUE #244: Exam mode configuration always enables Pitest and JaCoCo
PR #217 introduced an exam mode security guard which creates a copy of the configuration and discards the original one in order to prevent students from leaking meta tests or code checks during exams (in case the teacher forgets to remove them). This new configuration is supposed to be the same as the original, with the exception of meta tests and code checks. However, the current one has a bug: the `skipJacoco` and `skipPitest` methods aren't overridden, so they always return false, causing Pitest and JaCoCo to always be enabled during exams.

alexcojocaru2002:
Closes #244


martinmladenov (approved):
LGTM!

PR #268, opened 2023-09-19
Removed comments + added line highlight at position of keyword

alexcojocaru2002:
The aim of this Pull Request is to improve the [SourceCodeSecurityCheckStep](https://github.com/SERG-Delft/andy/blob/main/andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java) class. 

- [ ] Change name of the class
- [ ] Remove comments when checking code
- [ ] When having an error highlight the corresponding line and not line 1.

### DIFF
--- a/andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
+++ b/andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
@@ -12,6 +12,8 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
@@ -30,7 +32,6 @@ public class SourceCodeSecurityCheckStep implements ExecutionStep {
         }
 
         String code;
-
         try {
             code = Files.readString(Path.of(solutionFile.get(0).getPath()));
         } catch (IOException e) {
@@ -38,6 +39,7 @@ public class SourceCodeSecurityCheckStep implements ExecutionStep {
             return;
         }
 
+        code = removeComments(code);
         if (!checkPackageName(code, result)) return;
 
         if (!checkForKeywords(code, result)) return;
@@ -46,7 +48,7 @@ public class SourceCodeSecurityCheckStep implements ExecutionStep {
     private boolean checkPackageName(String code, ResultBuilder result) {
         Pattern pattern = Pattern.compile("^\\s*package\\s+delft\\s*;.*", Pattern.DOTALL);
         if (!pattern.matcher(code).find()) {
-            result.compilationSecurityFail("The package name of your solution must be \"delft\"");
+            result.compilationSecurityFail("The package name of your solution must be \"delft\"", null);
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


alexcojocaru2002:
@martinmladenov How should I rename the class ? Also should I write a separate test for removeComments method  ?

andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
```diff
@@ -90,14 +92,32 @@ private boolean checkForKeywords(String code, ResultBuilder result) {
...
+                        result.compilationSecurityFail(keywords.get(keyword), Optional.of(lineNumber+1));
+                        return false;
+                    }
             }
         }
 
         return true;
     }
 
+    public String removeComments(String code) {
```
mauricioaniche, line 109: Should this be a `public` method, or `private`?

andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
```diff
@@ -90,14 +92,32 @@ private boolean checkForKeywords(String code, ResultBuilder result) {
         );
         for (String keyword : keywords.keySet()) {
             if (code.contains(keyword)) {
-                result.compilationSecurityFail(keywords.get(keyword));
-                return false;
+                String[] lines = code.split("\\n");
+                // Split lines and search every line for keyword
+                for (int lineNumber = 0; lineNumber < lines.length; lineNumber++)
```
mauricioaniche, line 97: Can we have a test showing what this should do? :)

martinmladenov:
@alexcojocaru2002 
> @martinmladenov How should I rename the class ? Also should I write a separate test for removeComments method ?

Good question, I haven't thought about that. Maybe something like "SourceCodeCheckStep" would be suitable (but this could be misleading as people might think it has to do with code checks). If you think of anything more suitable, feel free to use that instead. @mauricioaniche Do you have any ideas?

mauricioaniche:
`SourceCodeCheckStep` looks good to me!

andy/src/main/java/nl/tudelft/cse1110/andy/result/ResultBuilder.java
```diff
@@ -78,9 +78,9 @@ public void compilationFail(List<Diagnostic<? extends JavaFileObject>> errors) {
         this.compilation = CompilationResult.compilationFail(compilationErrors);
     }
 
-    public void compilationSecurityFail(String message) {
+    public void compilationSecurityFail(String message, Optional<Integer> lineNumber) {
```
martinmladenov, line 81: You have an `Optional` here but you're passing `null` as an argument on line 51 of `SourceCodeSecurityCheckStep.java`. This leads to an exception.

I don't think it's a good idea to use an Optional as a method argument - it's way too easy for someone to accidentally pass `null` instead of `Optional.empty()`.

andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
alexcojocaru2002: Private 😓

andy/src/main/java/nl/tudelft/cse1110/andy/result/ResultBuilder.java
martinmladenov: Yeah, that seems like a good idea - you can overload the method:
- `compilationSecurityFail(String message, int lineNumber)`, which does the same as what you have now (except that the optional is replaced with a simple int)
- `compilationSecurityFail(String message)`, which calls `compilationSecurityFail(message, 1)`

andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
alexcojocaru2002: A bit of a stupid question but in order to test this should I write unit tests for the three check methods or write integration tests ?

PR #283, opened 2025-02-23
Update dependencies

martinmladenov:


andy/src/main/java/nl/tudelft/cse1110/andy/utils/JUnitUtils.java
```diff
@@ -6,10 +6,10 @@ public class JUnitUtils {
 
     public static String simplifyTestErrorMessage(TestExecutionSummary.Failure failure) {
```
fm2003, line 7: Would it be possible to have an integration test for this method? Same for `getFailingMethod`.

martinmladenov: There are already some integration tests covering this part:
https://github.com/SERG-Delft/andy/blob/22ce179a8789cac064c5cd2ac3e440be501ae143/andy/src/test/java/integration/JUnitTestsTest.java#L152-L183
https://github.com/SERG-Delft/andy/blob/22ce179a8789cac064c5cd2ac3e440be501ae143/andy/src/test/java/integration/JUnitTestsTest.java#L344-L354
https://github.com/SERG-Delft/andy/blob/22ce179a8789cac064c5cd2ac3e440be501ae143/andy/src/test/java/integration/JUnitTestsTest.java#L357-L369

PR #287, opened 2025-03-01
Add tests for parsing Unicode

### LINKED ISSUE #286: Unicode characters cause errors in Docker
Andy (or javac?) tries to parse Unicode characters as ascii and fails. \ufffd is the Unicode "replacement character", it's inserted when parsing a Unicode character fails. It appears that this issue occurs only in Docker.

Image

With the Maven plugin, it works fine.

Image

martinmladenov:
It turned out that the issue is not specific to the Docker image. It is dependent on the JDK version used; it does not occur in Maven if you have a newer version of javac on your system. In Java 18+, UTF-8 is the default encoding[^1]. In Java 17 (which the Docker image uses), it appears to be ASCII.

[^1]: https://openjdk.org/jeps/400

I fixed the issue by passing the `encoding UTF-8` argument to javac. When a zero-width space is present, a correct error is thrown:

```
Andy v1.2.0-387c684 (2025-03-01T14:13:01Z)

We could not compile the code. See the compilation errors below:
- line 18: illegal character: '\u200b'
        @Test
        void unicodeTest() {
-->         assertThat(Library.a()).​isEqualTo(1);
        }
    }

- line 18: not a statement

Andy took 0.2 seconds to assess your solution.
```

The zero-width space is before `isEqualTo`.

Previously, this was thrown:

```
We could not compile the code. See the compilation errors below:
- line 18: unmappable character (0xE2) for encoding US-ASCII
        @Test
        void unicodeTest() {
-->         assertThat(Library.a()).​isEqualTo(1);
        }
    }

- line 18: unmappable character (0x80) for encoding US-ASCII
- line 18: unmappable character (0x8B) for encoding US-ASCII
- line 18: illegal character: '\ufffd'
- line 18: not a statement
- line 18: illegal character: '\ufffd'
- line 18: illegal character: '\ufffd'

Andy took 0.2 seconds to assess your solution.
```

Fixes #286


martinmladenov:
@fm2003 Would it make sense to add custom handling for this compilation error to let students know that they have an invisible character on this line, or is the "illegal character" error thrown by the compiler sufficient?

martinmladenov:
It is not necessary to specify the javac argument if we update to a newer Java version (#289). Should we still add it?
(In any case, I think we should at the very least add the new tests).

fm2003:
> It is not necessary to specify the javac argument if we update to a newer Java version (#289). Should we still add it? (In any case, I think we should at the very least add the new tests).

Let's remove the javac argument and keep the new tests. 👍

fm2003:
> @fm2003 Would it make sense to add custom handling for this compilation error to let students know that they have an invisible character on this line, or is the "illegal character" error thrown by the compiler sufficient?

The Monaco editor now highlights invisible characters in source code. That in combination with an "illegal character" error should be sufficient for students to find out what the problem is.

PR #289, opened 2025-03-01
Update to Java 23

martinmladenov:
Updated to Java 23. I ran tests locally and everything works without issues.

I also tested Java 21 (LTS) in the `java-21` branch. It also works, but I think it would be best to jump directly to 23 and update it to the next LTS version, which will be released in September 2025.


fm2003:
Merged as part of #287.

PR #293, opened 2025-03-01
Block more methods and improve Selenium driver error message

### LINKED ISSUE #255: Change unsupported Selenium driver error message
Andy only supports the HtmlUnitDriver and doesn't support other drivers such as FirefoxDriver, ChromeDriver, etc. The error message currently says "Please use HtmlUnitDriver as Andy does not have a browser installed". This may be a bit confusing to students, as in exercises we always use a facade for the driver logic (`CSE1110.createDriver()`) so it might be better to change the error message to refer to this method instead. For example: "Please use CSE1110.createDriver() to instantiate a Selenium driver".

martinmladenov:
Resolves #255


andy/src/main/java/nl/tudelft/cse1110/andy/execution/step/SourceCodeSecurityCheckStep.java
```diff
@@ -38,46 +38,37 @@ public void execute(Context ctx, ResultBuilder result) {
...
-        }
-
-        return true;
-    }
-
     private boolean checkForKeywords(String code, ResultBuilder result) {
-        String reflectionMsg = "Using reflection in your code is not allowed";
-        String seleniumUnsupportedDriverMsg = "Please use HtmlUnitDriver as Andy does not have a browser installed";
+        String msg = "It is not allowed to use '%s' in your code";
+        String seleniumUnsupportedDriverMsg = "Please use CSE1110.createDriver() to instantiate a Selenium driver";
```
fm2003, line 46: This message seems rather specific. Maybe we can change it to: "Please use the provided method to instantiate a Selenium driver".

martinmladenov: Sure, I changed it

